PR #8, opened 2018-05-25
Added section on numerics

gusthoff:


courses/intro-to-ada/intro_to_ada.rst
```diff
@@ -8929,6 +8929,339 @@ any index in the file. Therefore, when changing the index, we need to make
...
+The standard library provides support for common numeric operations on
+floating-point types, as well as complex types and matrices. This section
+presents a brief introduction into the topic.
+
+Elementary Functions
+~~~~~~~~~~~~~~~~~~~~
+
+The :ada:`Ada.Numerics.Generic_Elementary_Functions` package provides
+common operations for floating-point types, such as square-root,
+logarithm, cosine, etc. Because it's a generic package, it needs to be
```
raph-amiard: The section is well written, but seems to be missing an important fact: nongeneric instantiations of the package exist for every built-in floating point type: `Ada.Numerics.Float_Elementary_Functions` is one example. They're pretty well hidden in the RM :) I think we should showcase those versions first! What do you think ?

PR #14, opened 2018-06-01
Initial commit of writing guidelines

gusthoff:


CONTRIBUTING.md
```diff
@@ -0,0 +1,48 @@
...
+=====================================
+
+Writing Guidelines
+------------------
+
+- Text must contain up to 79 characters per line.
+
+- Source-code examples should be complete, compilable, and free of errors.
+
+    - Valid exceptions for incomplete code are, for example, when the
```
raph-amiard: I actually disagree with this style, and it is not used in the intro-to-ada course to the best of my knowledge

CONTRIBUTING.md
```diff
@@ -0,0 +1,48 @@
...
+
+    - Available roles: `ada`, `c`, `cpp`.
+
+- References to other programming languages should be kept at a minimum
+  level.
+
+    - They should be used only when highlighting important differences for
+      common use-cases in widely-known programming languages.
+
+    - They should be part of admonitions, so that they don't disrupt the
```
raph-amiard: Full agreement, but specifiying the name of the admonition here would be great!

CONTRIBUTING.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Text must contain up to 79 characters per line.
+
+- Source-code examples should be complete, compilable, and free of errors.
+
+    - Valid exceptions for incomplete code are, for example, when the
+      writer wants to highlighting a specific source-code block from a
+      previous example.
+
+    - Erroneous source-code examples are acceptable when highlighting an
```
raph-amiard: I also disagree with this and the following paragraph. We have plenty of those in the intro-to-ada course; The fact that we have the code runner makes it fine IMHO. Of course we shouldn't have 75% of incorrect code examples, but that's common sense I think ?

For reference we have:

11 ada-expect-compile-error
24 ada-nocheck
3 ada-run-expect-failure
4 ada-syntax-only

Keeping in mind that amongst the ada-nocheck, probably some of them should be syntax-only or expect-failure. Thats 42 erroneous examples on 200.

The only thing that I think we could point out is that erroneous examples should be as short as possible.

CONTRIBUTING.md
gusthoff: It _is_ actually used in the course in some places. Just look for code blocks marked with `ada-nocheck`: some of them are referring to previous examples, highlighting one statement from that example and then saying something like "we could rewrite that statement in the previous example like this: [_incomplete code here containing single line_]". In my opinion, this is a valid exception to the rule; if we repeat the same code from the previous example once more just to change one single line, then it could be more confusing to the reader.

CONTRIBUTING.md
gusthoff: Now I'm lost: what exactly are you disagreeing with? You told me once that source-code examples should be complete and correct; if we need to mention pitfalls / potential errors, we should not write source-code examples for that, but rather just mention the pitfalls in the text. To be more specific, we discussed that topic during this code review:

https://github.com/raph-amiard/intro-to-ada/pull/9

For example, this is one remark you made:

> I think you should not make people read 25 lines of code that is wrong. You should not assume that people have read everything around it.

Also, we discussed that topic again in a call afterwards and agreed that mentioning pitfalls in the text would be enough (no need for erroneous source-code examples).

What I'm trying to say is that the rules I'm writing here just reflects my mindset while writing lessons for this course. If you look at the stuff I've written, you'll see that I used `ada-expect-compile-error` just once and didn't use `ada-run-expect-failure` at all. The single instance where I used `ada-syntax-only` was when mentioning automatic generated bindings for C code (Ada specification file, no runnable code). Also, most of the 10 instances of `ada-nocheck` I used in the course where just incomplete code referring to small changes in statements from previous examples (as I explained in the previous comment above). In summary, my goal was to write complete, correct source-code examples...

If you want to rephrase this whole paragraph, however, it's fine with me. Just make sure to explain in details what your expectations are.

gusthoff:
In general, I'd prefer you just merge this PR as an initial commit and then rewrite the rules to your liking. :-)

raph-amiard:
OK done :)

PR #15, opened 2018-06-02
Pass on the first chapter of the SPARK intro

setton:
And replace some non-ASCII characters.


raph-amiard:
Hey there! I'm not sure about using 1 code block to regroup code that goes together, I would like us to discuss this first before approving this review :)

PR #19, opened 2018-06-05
Cleanup for first 3 sections of SPARK course

yannickmoy:
At this time, there is no need anymore to emphasize the disctinction
between previous SPARK versions and current SPARK version.

### DIFF
--- a/courses/intro-to-spark/book/01_Overview.rst
+++ b/courses/intro-to-spark/book/01_Overview.rst
@@ -244,9 +244,10 @@ No aliasing of names
 
 Another restriction imposed in the SPARK subset concerns aliasing. We say that
 two names are aliased if they refer to the same object. Since access types
-(pointers in Ada) are not allowed in SPARK, aliasing can only occur as part of
-the parameter passing in a procedure call. As a consequence, when a procedure
-is called, SPARK makes sure that no :ada:`out` or :ada:`in out` parameter is
+(`pointers <https://en.m.wikipedia.org/wiki/Pointer_(computer_programming)>`_
+in Ada) are not allowed in SPARK, aliasing can only occur as part of the
+parameter passing in a procedure call. As a consequence, when a procedure is
+called, SPARK makes sure that no :ada:`out` or :ada:`in out` parameter is
 aliased with either another parameter of the procedure or a global variable
 updated in the procedure's body.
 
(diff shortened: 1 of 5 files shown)


courses/intro-to-spark/book/01_Overview.rst
```diff
@@ -228,48 +232,42 @@ attempts to update the global variable ``Call_Count``.
...
-Another restriction imposed in the SPARK subset concerns aliasing. We say
-that two names are aliased if they refer to the same object. Since access
-types are not allowed in SPARK, aliasing can only occur as part of the
-parameter passing in a procedure call. As a consequence, when a procedure
-is called, SPARK makes sure that no :ada:`out` or :ada:`in out` parameter
-is aliased with either an other parameter of the procedure or a global
-variable updated in the procedure's body.
+Another restriction imposed in the SPARK subset concerns aliasing. We say that
+two names are aliased if they refer to the same object. Since access types
+(pointers in Ada) are not allowed in SPARK, aliasing can only occur as part of
```
gusthoff: In general, we're avoiding terminologies from other languages. If you fear, however, that the concept of access types is not clear here, maybe a link to the _Intro to Ada_ course would be nice.

courses/intro-to-spark/book/02_Flow_Analysis.rst
```diff
@@ -50,30 +50,30 @@ prior to entering the loop. As a consequence, the value read in the
 condition of the if statement may be uninitialized. Flow analysis will
 detect and report this error.
 
-.. code:: ada
+.. code-block:: ada
+
+    package Show_Uninitialized is
```
gusthoff: Just for my understanding: isn't the `SPARK_Mode` needed here (and also in other examples below)?

yannickmoy: I explained why I did that on the ticket, using SPARK_Mode is redundant with the global configuration pragma that we use throughout (set up with the generated project file). I think it's cleaner to avoid adding SPARK_Mode everywhere.

courses/intro-to-spark/book/01_Overview.rst
yannickmoy: since we're also targeting people who don't know Ada, I'd rather be clear about what access types are. I'll add a link to the intro-to-ada section.

raph-amiard: @gusthoff I think "pointer" is an acceptable terminology. It is much more general than C/C++, in fact Ada is pretty much the odd duck here.

@yannickmoy a link to https://en.m.wikipedia.org/wiki/Pointer_(computer_programming) on this link would still be an appreciated addition there I think.

PR #20, opened 2018-06-05
Fixes & sphinx build for SPARK course

raph-amiard:


raph-amiard:
@yannickmoy should be fixed!

PR #24, opened 2018-06-06
Move testing to the engine/ directory

setton:
And a simple Makefile to drive the tests.


raph-amiard:
I think this PR is redundant because the two commits were already submitted as part of #25, so I'll close it!

PR #25, opened 2018-06-06
transition the Ada course to code widgets

setton:


courses/old-adacoreu-content/01_Overview/book/01_Your_First_Ada_Program.rst
```diff
@@ -1,223 +0,0 @@
-Lesson 1: Your First Ada Program
```
gusthoff: Is there a reason for removing these transcriptions? I thought there was a common understanding that this repository would be used for all learning material (old and new)... Is this stuff being moved somewhere else?

PR #28, opened 2018-06-12
Infrastructure fixes and improvements

raph-amiard:


courses/index.rst
```diff
@@ -0,0 +1,27 @@
...
+.. raw:: html
+
+    <hr>
+    <div id="content_blocks">
+
+.. toctree::
+   :maxdepth: 2
+
+   Intro to ada <intro-to-ada/index>
+   Intro to ada <intro-to-spark/index>
```
yannickmoy: Intro to Ada
Intro to SPARK

courses/intro-to-ada/index.rst
```diff
@@ -0,0 +1,20 @@
...
+Introduction to Ada
+===================
+
+.. toctree::
+    :maxdepth: 2
+    :caption: Contents:
+
+    intro_to_ada
+
+Indices and tables
```
yannickmoy: Do you need an index here, since both course are part of an enclosing one?
Are these local index only reporting on local stuff?

PR #39, opened 2018-06-21
take into account feedback from Claire & Boris

yannickmoy:
minor changes and fixes

### DIFF
--- a/courses/intro-to-spark/book/01_Overview.rst
+++ b/courses/intro-to-spark/book/01_Overview.rst
@@ -254,8 +254,9 @@ while ``Incr_And_Log`` is not as it attempts to update the global variable
 No aliasing of names
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Another restriction imposed in the SPARK subset concerns aliasing. There are
-two reasons to forbid aliasing in SPARK:
+Another restriction imposed in the SPARK subset concerns aliasing. We say that
+two names are aliased if they refer to the same object. There are two reasons
+to forbid aliasing in SPARK:
 
 - First, it makes verification more difficult as it requires taking into
   account the fact that updates to two variables with different names may in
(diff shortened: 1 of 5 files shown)


courses/intro-to-spark/book/01_Overview.rst
```diff
@@ -254,8 +254,9 @@ while ``Incr_And_Log`` is not as it attempts to update the global variable
 No aliasing of names
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Another restriction imposed in the SPARK subset concerns aliasing. There are
-two reasons to forbid aliasing in SPARK:
+Another restriction imposed in the SPARK subset concerns aliasing. We say that
+two names are aliased if they refer to the same object. There are two reasons
```
gusthoff: Maybe we could also link to a Wikipedia page? https://en.wikipedia.org/wiki/Aliasing_(computing)
Just an idea, feel free to drop it.

courses/intro-to-spark/book/05_Proof_Of_Functional_Correctness.rst
```diff
@@ -969,10 +969,50 @@ condition` stating what part of the array has not been modified so far:
...
+
+    end Show_Map;
+
+    package body Show_Map is
+
+       procedure Map (A : in out Nat_Array) is
+          A_I : constant Nat_Array := A with Ghost;
+       begin
+          for K in A'Range loop
+             if K /= A'First then
```
gusthoff: I'd rather adapt the range `for K in` to exclude the first element instead of using this `if` inside a loop. However, I'm maybe missing the point of this code example.

yannickmoy: you can't without modifying elsewhere the example, as you'll need to write "K+1
" which by itself may overflow if K = Integer'Last. :-)

You can see that if you try in the live version.

yannickmoy:
thx Gustavo

courses/intro-to-spark/book/05_Proof_Of_Functional_Correctness.rst
gusthoff: You mean, modifying to something like this?

```
      if A'First /= Positive'Last then
        for K in A'First + 1 .. A'Last loop
```

yannickmoy: yes, that would work, but I don't find it more readable, do you?

gusthoff: I agree, readability is the same. However, that modification makes the code more efficient. Since I care about efficiency, seeing an avoidable `if` inside a `loop` sends shivers down my spine. ;-)

Anyway, no need to change the code if you don't want to.

yannickmoy: I'm sure the compiler optimizations will figure it out. :-)
Feel free to modify it, at this stage I'm inclined to leave it as it is.

PR #40, opened 2018-06-26
Richard's editing so far

RichardKenner:


RichardKenner:
>   +    - When bounds or subtype doesn't match or, in general, any violation of
> 
> Shouldn't it rather be "When bounds or subtype don't match" (i.e.: "when [they]
> don't match")?

No.  For "A or B", you take the plural/no plural from A and B.  With
"and", it's always plural, but not for "or".  If they mismatch,
convention is to use the latter.  So it's technically correct, but
it's hard to read with the difference on both sides of the "or" so
I'll fix it.  I think the fix is to remove "or subtype" because it's
not clear what it refers to here.

> I agree that the original phrase is not great, but this correction sounds
> confusing to me:
> "In C, we simply declare the function [...], just like we would if it were a
> function in C."

Agreed.  I struggled with that.  Let me try again.

    On the C side, we do the same as we would do if the function were
    written in C: simply declare it using the :c:`extern` keyword.  For
    example:

> Missing word: "The [?] does the best".

"Compiler"

> This sounds confusing: "[...] as the type of one of more pointers." Did you
> mean "of one or more [...]"?

Yes.

> I still prefer "a developer" or "developers" instead of "a human". But I
> understand you want to highlight the contrast Human vs. Machine.

Exactly.  This is one place where I think "human" is appropriate.

PR #42, opened 2018-06-29
Edit of standard_library

RichardKenner:


courses/intro-to-ada/chapters/standard_library.rst
```diff
@@ -1748,18 +1733,17 @@ the content of strings:
...
+All these operations are available both as functions or procedures.
+Functions create a new string but procedures perform the operations in
+place. The procedure will raise an exception if the constraints of the
+string are not satisfied. For example, if we have a string ``S`` containing
+10 characters, inserting a string with two characters (e.g. ``"!!"``) into
+it produces a string containing 12 characters. Since it has a fixed length,
+we can't increase its size. One possible solution in this case is to
+specify that truncation should be applied while inserting the substring.
+This keeps the length of ``S`` fixed. Let's see an example that makes use
+of both function and procedure version of ``Insert``, ``Overwrite``, and
```
setton: * versions 
? (not sure)

courses/intro-to-ada/chapters/standard_library.rst
```diff
@@ -1859,31 +1841,30 @@ time:
...
 Bounded strings
 ^^^^^^^^^^^^^^^
 
 Bounded strings are defined in the
 :ada:`Ada.Strings.Bounded.Generic_Bounded_Length` package. Because this is
-a generic package, we need to instantiate it first and set the maximum
-length of the bounded strings. We can then declare bounded strings using
-the ``Bounded_String`` type.
+a generic package, you need to instantiate it and set the maximum length of
+the bounded strings you we can use it. you can then declare bounded strings
```
setton: "you we can use it" <- too many words here :)

courses/intro-to-ada/chapters/standard_library.rst
```diff
@@ -2310,15 +2287,14 @@ replace the instances of the :ada:`Ada.Sequential_IO` package by the
...
     end Show_Dir_Float_IO;
 
-The main difference between sequential I/O and direct I/O is that direct
-I/O allows for accessing any position in the file. Also, it doesn't offer
-an option to append information to a file. Instead, it has an
-``Inout_File`` mode that allows for reading and writing to a file using
-the same ``File_Type`` element.
+Unlike sequential I/O, direct I/O allows you to access any position in the
+file. However, it doesn't offer an option to append information to a
+file. Instead, it provides an ``Inout_File`` mode allow reading and writing
```
setton: allowing?

RichardKenner:
>   -  are or present or for erroneous execution.
>   +  exist or for case of erroneous execution.
> 
> can you say "for case of" ? I would say "in case of", no?

That's indeed a garble.  I try to carefully review changes I made in
the last pass, but missed this one.  That whole thing doesn't read well,
so let me try again.

> Length (V)

Yup.  I neither introduced or caught that.  ;-)

> * is to use

Yup.


> ... is no longer present ?

Oops ...

> *the :ada:String type

I could go both ways on that, but let's use "the".

> *considered

Got it.

> "if we'd passed" ... this contraction looks funny to me, perhaps because I'm
> not a native speaker. Would "if we had passed" work better? I suppose "had we
> passed" is too fancy?

It sounds OK to me, but I'm not the one who counts here.  It's a contraction
for "we had".  Probably simplest to just remove the "had":

    However, if we passed a date in the future, the program would wait
    until that specific date and time arrived.

>   +In the examples above, we were delaying to a specific date and time.  Just
>   +like we saw in the tasking chapter, we could instead delay relative to the
> 
> not "relatively" ?

No, but let me rewrite that.  "delay relative to" is a quite common
construction, but I think adding "specify the delay relative to" may
make it clearer, so I did that.

> * versions

Yes.  "both" and "and" point to the plural.

> "you we can use it" <- too many words here :)

Garble fixed.

> allowing?

Yes.

Thanks very much for the pass here!  I thought that three passes would
be enough to get this stuff out of it, but I guess not.

PR #43, opened 2018-07-03
Review the runnable examples

setton:
Add "no_button" where needed. Some other small adjustments.
Deactivate buttons on the chapter on interfacing with C, since
the widget/gnatchop do not support C files for now.


courses/intro-to-ada/chapters/standard_library.rst
```diff
@@ -1553,7 +1553,7 @@ loop. For example:
 
           X : Long_Long_Float;
        begin
-          for I in 0 .. 5_000_000 loop
+          for I in 0 .. 1_000_000 loop
```
gusthoff: I prefer the original value (because the relation "CPU time vs. runtime" is more explicit), but I'm OK with changing it.

setton: it was a bit slow on the cloud machine...

PR #48, opened 2018-07-06
TOC: correct max. depth

gusthoff:


Robert-Tice:
Do we want the toc page for each course to show all sub content? Or just top subject headings? We could remove the max depth directive if we just want everything to show up.

PR #51, opened 2018-07-09
Use more complex example code on first page, with Run button too

yannickmoy:


Fabien-Chouteau:
Maybe a bit "cheesy" but I propose this:

```Ada
with Ada.Text_IO; use Ada.Text_IO;

procedure Main is
   
   subtype Alphabet is Character range 'A' .. 'Z';

begin

   Put_Line ("Learning Ada from " & Alphabet'First & " to " & Alphabet'Last);

end Main;
```

yannickmoy:
feel free to push that other example, but then remove "Prove" button as there is nothing to prove in that snippet. And add run_button.

Robert-Tice:
The example accesses Ada.Command_Line which the user can't utilize anyway with the infrastructure so I think I like @Fabien-Chouteau example.

raph-amiard:
Fabien can you take care of it ?

raph-amiard:
@Fabien-Chouteau can you take care of it ?

PR #53, opened 2018-07-11
Adding publish-staging target to Makefile.

Robert-Tice:


engine/Makefile
```diff
@@ -22,11 +22,41 @@ test_spark_intro:
...
+	# remove all files from the repo to get rid of stale copies
+	@rm -rf *
+
+	# reset files that are needed that won't be copied from the build
+	@git reset CNAME
+
+	# copy build files to learn-staging-html-pages
+	@cp -R $(BUILDDIR)/html/. .
+
+	# commit and push new changes
```
raph-amiard: Is that on purpose ? (commented lines ?)

Robert-Tice: I seem to have commited a testing version of the Makefile rather than the final version. Will fix now.

Robert-Tice:
The checks failed because Makefiles require tabs instead of spaces and yeah....

PR #111, opened 2018-07-27
Added complete source-code examples to Advanced SPARK course

gusthoff:


courses/advanced-spark/chapters/concurrency.rst
```diff
@@ -105,18 +105,25 @@ Concurrency – A trivial example
 
 .. code:: ada
 
-    task type T;
+    procedure Show_Trivial_Task is
+       task type T;
```
yannickmoy: This cannot work with SPARK. You will get a SPARK error, as SPARK follows Ravenscar restrictions so a task has to be declared at library-level (not inside a subprogram). And in addition I think we should not show examples of procedures that do nothing just being scopes for code snippets. I changed all such examples in the basic SPARK course. BTW, the 2 pragmas described in SPARK UG (http://docs.adacore.com/spark2014-docs/html/ug/en/source/concurrency.html) should be set in a pragma configuration file (the one with currently "pragma SPARK_Mode(On);" used for all SPARK examples) or GNATprove will reject all the concurrent code.

courses/advanced-spark/chapters/ghost_code.rst
```diff
@@ -61,41 +71,68 @@ Ghost variables – aka auxiliary variables
 
 .. code:: ada
 
-    Data : Data_Array;
-    Free : Natural;
+    procedure Show_Ghost_Variable is
```
yannickmoy: why use a procedure here? This looks artificial. I'd prefer a package spec+body, or a package to define types and a top-level procedure Alloc.

courses/advanced-spark/chapters/subprogram_contracts.rst
```diff
@@ -88,26 +94,49 @@ Precondition
 
 .. code:: ada
 
-    function Sqrt (X : Float) return Float with
-      Pre => X >= 0.0 or else raise Argument_Error;
+    package Show_Precondition is
```
yannickmoy: This looks exactly like the previous Show_Dynamic_Behavior.

courses/advanced-spark/chapters/concurrency.rst
gusthoff: Yes, you're absolutely right. (I copied the wrong example from my local notebook, even though I had the correct one using packages right below. Shame on me!)

courses/advanced-spark/chapters/subprogram_contracts.rst
gusthoff: And it is. :-) It was like that in the original version of the slides. 

I didn't change the original code examples. but rather just created a "complete" version based on them.

PR #120, opened 2018-08-21
Topic/richard

RichardKenner:


RichardKenner:
>   +are available through the buttons labelled:
> 
> labeled ?

Actually, either is an acceptable spelling.  I prefer the double l, but
for no good reason.

>   +SPARK. while Function ``Incr`` might be legal if it doesn't modify any
> 
> SPARK. -> SPARK,

Fixed.

>   +here instead of an error giving you the possibility of justifying the
> 
> suggest adding a comma after 'an error'

Yup.

RichardKenner:
> Small editorial remark: there are two whitespaces instead of one here and
> there.

Not "here and there": it's after a period.  Old-school, I guess, but it
doesn't matter for .rst.

>   +Ineffective statements are different than dead code: they're executed, and
>   +often even modify the value of variables, but have no effect on any of the
>   +subprogram's visible outputs: parameters, global variables or the function
>   +result. Ineffective statements, like unused variables, should be avoided,
> 
> The comma after "avoided" breaks the statement unnecessarily. Is it really
> correct?

Yes, because it disambiguates between

     "avoided in this case, because"

and 

"avoided, in this case because"

which have two different meanings.  We mean the latter.

However, I'm trying not to write hard-to-parse sentences here because they
distract from the material of the course, so I'll change this.  I don't really
see the need to refer to undefined variables here, so I changed this to:

Ineffective statements should be avoided because they make the code
less readable and more difficult to maintain.

> Maybe "by specifying them"?

Yes.

>   +subprogram returns, it check that global and parameter outputs were
> 
> *checks

Got it.

>   +the ability to accurately detect of reads of uninitialized data.
> 
> Shouldn't we remove the "of" in "to accurately detect of reads"?

Yes.

> A backtick is missing for the "A" at the end of the line.

Fixed.

>   +analysis cannot guaranty that ``Result`` is always initialized at the call
> 
> *guarantee

Yup.

>   +nothing to be uninitialized.
> 
> *initialized

Fixed.

PR #124, opened 2018-09-10
Edits to book #5

RichardKenner:


courses/intro-to-spark/book/05_Proof_Of_Functional_Correctness.rst
```diff
@@ -1085,19 +1107,19 @@ contracts.
...
     end Ring_Buffer;
 
-This is correct as ``Get_Model`` is used only in contracts. Note that calls to
-``Get_Model`` cause copies of the buffer's content, which is not
-efficient. This is fine because ``Get_Model`` is only used for verification,
-not in the final production code. This is enforced by making it a ghost
-function and producing the final production code with appropriate compiler
-switches (not using ``-gnata``) that ensure that assertions are ignored.
+This is correct: ``Get_Model`` is used only in contracts.  Calls to
+``Get_Model`` makes copies of the buffer's contents, which isn't efficient,
```
gusthoff: *_make_

RichardKenner:
> Minor suggestion: I'd write to "to ensure that these entities can't
> [...]" to keep this part consistent with the "or that they're
> completely [...]" part, which comes later in the same phrase.

Agreed.

> This requires rephrasing: "the the compiler flags the use of [...]"

Duplicated word deleted.

> Editorial comment: newlines in the ":ada:" role usually cause issues
> when rendering the ReST file to its final format. At least, I had
> this issue when using pandoc to generate a PDF file: the ":ada:"
> roles containing newlines were not rendered correctly. Therefore,
> I'd suggest to move this to the next line.

Good point.

> Editorial comment: again, please avoid newlines within the ":ada:" role.

Done and I checked all of them.  This happened due to emacs autofill.

>   +Some intermediate assertions may not be proved by GNATprove
>   either because +it's missing some information or because the
>   amount of information +available is confusing to it. You can
>   verify these remaining assertions by
> 
> Editorial suggestion: because the phrase is long, I'd change
> "confusing to it" to "confusing to GNATprove".

I just deleted "to it".  I think that's implied and makes the sentence
even easier to read.

>      :ada'Old:
> 
> (misplaced colon at the end)

Fixed.

>   +represent this value and writing an assertion in both branches of an
>   +:ada:`if` statement that repeat the postcondition, but using ``X_Init``.
> 
> *repeats

Got it.

> +don't create new elements, that is, that all the elements present
> +in ``A``
> 
> *doesn't create

Done.

> Editorial suggestion: the phrase starting with "This property isn't
> [...]" is very long. It's also difficult to understand without
> reading it twice (in my opinion, of course), so I'd suggest
> splitting it into two phrases.

Yup.  Done.

>   +In the example below, calling ``Prove_P`` with ``X`` as an
>   +operand only +adds ``P (X)`` to the proof context instead of the
>   +larger set of assertions +required to verify it. In addition, the
>   +proof of ``P`` need only be done
> 
> Question: do we say "need only" or should it be "needs only"?

"need only": the proof need only 

> *let's look

Done.  And I checked for more.

>   +This is correct: ``Get_Model`` is used only in contracts.  Calls to
>   +``Get_Model`` makes copies of the buffer's contents, which isn't
>   +efficient,
> 
> *make

Yup.

PR #132, opened 2018-09-28
Adaptations to Sphinx infrastructure

gusthoff:


engine/travis_script.sh
```diff
@@ -6,4 +6,4 @@ which gcc
 which gnatchop
 which gprbuild
 
-make tests html
+make SPHINXOPTS="-W" tests html
```
Robert-Tice: The build test failed with this option on a warning in the Ada for CPP Java book. That book might require a bit of syntax cleanup before this option will allow the automated tests to complete.

PR #134, opened 2018-09-28
Sphinx: added Graphviz to list of extensions

gusthoff:


.travis.yml
```diff
@@ -17,6 +17,9 @@ cache:
   - $HOME/build_tools
   - $HOME/.cache/pip
 
+before_install:
+ - "sudo apt-get install graphviz"
```
Robert-Tice: Is this required? Can we just add graphviz to the requirements.txt file?

gusthoff: But "requirements.txt" is only for Python pip, right?  This is the actual Graphviz for the Linux machine.

Robert-Tice: O I see. It requires the installation for rendering and the python package doesn't include the full install.

PR #140, opened 2018-10-19
Compiled list of material that could be reuse

gusthoff:


gusthoff:
I've moved the SPARK books to the list of skipped items.

PR #156, opened 2018-11-09
Added section on type-based security

gusthoff:
Reusing Gems 82 and 83:
- https://www.adacore.com/gems/gem-82
- https://www.adacore.com/gems/gem-83

### DIFF
--- a/courses/advanced-ada/chapters/strong_typing.rst
+++ b/courses/advanced-ada/chapters/strong_typing.rst
@@ -16,6 +16,344 @@ Strong typing
 In this chapter, we discuss the advantages of strong typing and how it can
 be used to avoid common implementation and maintenance issues.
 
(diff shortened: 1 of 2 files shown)


courses/advanced-ada/chapters/strong_typing.rst
```diff
@@ -16,6 +16,344 @@ Strong typing
...
+       begin
+          return 1;
+       end Trusted_1;
+
+       function Trusted_100 return Trusted_Value is
+       begin
+          return 100;
+       end Trusted_100;
+
+       function "+" (V, W : Trusted_Value) return Trusted_Value is
```
yannickmoy: not sure you need to provide an explicit operation here, it should be available already as Trusted_Value is derived from Integer.

PR #171, opened 2018-12-01
Generalized call to method of derived tagged type

gusthoff:


courses/advanced-ada/chapters/generics.rst
```diff
@@ -1221,9 +1221,11 @@ the average of all elements.
     begin
        for I in A'Range loop
           A (I).Set (1.0);
-       end loop;
 
-       T2 (A (2).all).Set_Ext (3.0);
+          if A (I).all in T2'Class then
+             T2 (A (I).all).Set_Ext (3.0);
```
yannickmoy: here you're doing a non-dispatching call to Set_Ext. Possibly better to convert to T2'Class in order to have Set_Ext be dispatching no?

PR #178, opened 2018-12-26
Added section on factory functions

gusthoff:


courses/advanced-ada/chapters/design_patterns.rst
```diff
@@ -0,0 +1,960 @@
...
+       begin
+          Container.Rep.Top_Index := 0;
+       end Clear;
+
+       procedure Assign
+         (Target : in out Stack;
+          Source : Stacks.Stack'Class)
+       is
+          --  C : Stacks.Cursor'Class := Source.Top_Cursor;
+          C : Stacks.Cursor'Class := Top_Cursor (Source'Unchecked_Access);
```
gusthoff: Ideally, we should remove this `'Unchecked_Access`.

raph-amiard: Two ideas:

1. Try calling via dot notation: `Source.Top_Cursor`
2. Make Top_Cursor not take an access (I'm not sure why it needs to anyway)

gusthoff: Option 1 is actually the commented-out code, so that doesn't work. I suspect option 2 doesn't work either because the access type is used in the implementation of Top_Cursor function (``return Cursor'(Container, R.Top_Index);``). You can find it at line 843.

gusthoff: I've just tried the third option (putting the package instantiations into a separate package and used that in the main application). Unfortunately, I still get the same ``accessibility check failed`` error at run time.

PR #186, opened 2019-01-18
Initial commit of blog-post as learning material

gusthoff:
This is a proof-of-concept, no detailed review is required.

The idea is to evaluate publishing of blog-posts integrally in the learning website and adding source-code snippets that can be executed / proved by the reader.


gusthoff:
Yes, I agree with your comments. They are all valid and should be addressed before publishing this post. However, this is actually just the initial commit of an experiment / proof-of-concept. That's also the reason why the file is stored in the "internal" folder. Currently, there's no plan to publish this post.

yannickmoy:
OK, since you know well the challenges of such an integration, good with me!

yannickmoy:
btw, do you want me to merge after approving, of should I leave it to you?

gusthoff:
Thanks!

PR #196, opened 2019-02-08
Fix use of widget on SPARK book

yannickmoy:


books/SPARK_for_the_MISRA_C_Developer/chapters/03_syntactic_guarantees.rst
```diff
@@ -1,4 +1,4 @@
-:code-config:`run_button=False;prove_button=False;accumulate_code=False`
+:code-config:`run_button=True;prove_button=False;accumulate_code=False`
```
setton: this does not work, it will add "Run" buttons on the "package sign_domain" and "function absval" snippets, which don't have a main.

books/SPARK_for_the_MISRA_C_Developer/chapters/04_strong_typing.rst
```diff
@@ -1,4 +1,4 @@
-:code-config:`accumulate_code=False;run_button=True`
+:code-config:`run_button=True;prove_button=False;accumulate_code=False`
```
setton: same as above, if you do this, you'll end up with "Run" buttons for snippets that have no main

books/SPARK_for_the_MISRA_C_Developer/chapters/03_syntactic_guarantees.rst
yannickmoy: yes, and in that case it will just compile the code no? I'd like these snippets to be compiled instead of analyzed, so a run button is all I need?

books/SPARK_for_the_MISRA_C_Developer/chapters/04_strong_typing.rst
yannickmoy: yes, which corresponds to a compilation of those snippets, excactly what I want

PR #197, opened 2019-02-12
Complete review by Ben

yannickmoy:


books/SPARK_for_the_MISRA_C_Developer/chapters/03_syntactic_guarantees.rst
```diff
@@ -126,8 +126,8 @@ Handling Function Parameters
```
raph-amiard: Unrelated to Ben's review, but it's a bit weird to me that this is in the "syntactic" section of the book. I'd expect it to be in the semantic one, because the check that Ada/SPARK does is a semantic one, not a syntactic one (unlike for function returns where function and procedure calls are really treated differently by the parser)

raph-amiard: Reviewing the rest of the sections (such as the case statement), it seems to me your are mixing basic syntactic *and* semantic guarantees in this section. The simplest solution for me would be to rename the section into something like "Basic syntactic and semantic guarantees"

yannickmoy: for C the MISRA C rules are all syntactic, but indeed the alternative shown for SPARK go beyond the syntactic guarantees. I'd rather keep it this way for now, as we need to push the book to the publisher today

raph-amiard: Well I disagree. MISRA C Directive 4.7: "If a function returns error
information, then that error information shall be tested" is not purely syntactic. But it's a nit anyway I guess, do what you feel is best :)

yannickmoy: I agree with you here. I feel the "syntactic" main focus justifies the chapter title. OK for your review then?

PR #214, opened 2019-04-01
Draft version of book: "Intro to GNAT toolchain"

gusthoff:


Fabien-Chouteau:
Hi @gusthoff ,

Thanks for the great work :)

I have two comments:
 - I think that the code expansion and target dependent information are far too advanced and corner cases for an "intro to GNAT".

 - Since we already have a `Getting Started with GNAT Toolchain` course, this one seems like it should be merge in the existing getting started.

PR #239, opened 2019-06-28
Labs: draft version of exercises for "Modular Programming" chapter

gusthoff:


Robert-Tice:
There is a new format for the Lab IO block that has been introduced in the engine. The details of the change and the new format can be found in issue #242

PR #241, opened 2019-07-01
Fix leading whitespaces in lab expected output

Fabien-Chouteau:


Robert-Tice:
This will be fixed as part of issue #242

PR #268, opened 2019-08-30
Adding section on child and nested packages

gusthoff:


courses/intro-to-ada/chapters/modular_programming.rst
```diff
@@ -237,3 +239,434 @@ This example shows how :ada:`Last_Increment` is used indirectly:
        R := Increment_By (I);
        Display_Update_Values;
     end Main;
+
+Child and nested packages
+-------------------------
+
+Packages can be used to create hierarchies. We achieve this by using child
+or nested packages, which are both subunits of a parent package. One example
```
frank-at-adacore: I think saying "are both subunits" is not clear - internal and external visibility issues are very different. In my opinion, child packages generally extend the capability of the parent, whereas nested packages generally modularize the enclosing package. Your wording implies they could be used similarly

courses/intro-to-ada/chapters/modular_programming.rst
```diff
@@ -237,3 +239,434 @@ This example shows how :ada:`Last_Increment` is used indirectly:
...
+
+    procedure Main is
+       use Week.Nested;
+    begin
+       Put_Line ("First day of the week is " & Get_First_Of_Week);
+    end Main;
+
+Note that we cannot access the :ada:`Week.Nested` package directly using
+:ada:`with Week.Nested` because :ada:`Nested` is actually part of :ada:`Week`,
+not  a separate subunit. We can, however, still write :ada:`use Week.Nested`
```
frank-at-adacore: 1) extra space between "not" and "a" (not sure if it matters :)
2) "separate" is an Ada word - I would say  "an independent subunit"

sttaft:
The term "subunit" has a technical meaning in Ada, and should probably not
be used otherwise.  In Ada, a "subunit" is something introduced with a
"separate... " clause, and is associated with a subunit "body-stub" in the
parent unit.  See RM 10.1.3.

An alternative way to phrase this might be:

"... We achieve this using child or nested packages, which are both
logically contained within their parent package. ..."

-Tuck

sttaft:
The term "library-level" is a bit confusing.  I think you mean "library
unit" packages (or simply "library packages").  And I presume you mean to
include child packages in your term "library level."  In fact, everything
declared immediately inside a library unit package is also defined to be
"library level" in Ada (see RM 3.10.2(22)).

Your friendly language lawyer at work! ;-)

-Tuck

sttaft:
He does... ;-)
-Tuck

PR #279, opened 2019-10-07
updated year

MatrixMike:
I wanted to contribute to this great source of info and here is a test of the update system for me. 
Actually I also want to know, but do not yet know hoe else to tell you, that editing the labs on an iPad is tricky if you want to add "  as some other character is generated which the compiler rejects. 
My underlying reason for the latest enthusiasm is to get Ada on the BBC micro:bit - so far the instructions that I have seen are a bit short on explanations. I have nearly got it to work but it was a pain.
All the best
Mike


gusthoff:
Hello @MatrixMike ,

Thanks for your contribution, we need you to sign the Contributor License Agreement.

We've tested the issue you reported (editing the labs on an iPad). However, we weren't able to reproduce it. Maybe you have an alternative keyboard layout that generate those quotes?

PR #280, opened 2019-10-07
Fix invalid Java code example

### LINKED ISSUE #273: Incorrect Java example
The Java example in https://github.com/AdaCore/learn/blob/master/courses/Ada_For_The_CPP_Java_Developer/chapters/06_Functions_and_Procedures.rst#general-form is not valid Java code, because in Java everything must be in a `class`.

One possible fix would be to move `Func` into `class ProcData`.  Perhaps it would make sense to make it `static` as well.

Heziode:
Fix #273


raph-amiard:
@Heziode Thanks a lot for your contribution!

PR #281, opened 2019-10-12
Labs: draft version of exercises for "Object-oriented programming" ch…

gusthoff:
…apter

### DIFF
--- a/labs/intro-to-ada/chapters/object_oriented_programming.rst
+++ b/labs/intro-to-ada/chapters/object_oriented_programming.rst
@@ -1,0 +1,373 @@
+:orphan:
+
+Object-oriented programming
+===========================
+
+:code-config:`reset_accumulator=True;accumulate_code=False`
+
+.. include:: <isopub.txt>
+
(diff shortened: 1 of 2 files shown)


labs/intro-to-ada/chapters/object_oriented_programming.rst
```diff
@@ -0,0 +1,373 @@
...
+
+Simple type extension
+---------------------
+
+In this exercise, you'll work with record types containing numeric
+components. The first step is to create a record type :ada:`T_Float` with
+a floating-point component in the :ada:`Type_Extensions` package below.
+Next, you create a record type :ada:`T_Mixed` that extends the
+:ada:`T_Float` type and adds an integer component. In case of the
+:ada:`T_Mixed` type, the goal is to have both components *synchronized*:
```
frank-at-adacore: Not sure if this is just how things are done in AdaCoreU, or if this is a problem.
The "goal" is never verified - the only actual check done by main is the class derivation. Put_line could print out empty strings and the test would pass.

PR #285, opened 2019-11-02
Labs: draft version of exercises for "Standard Library" chapter

gusthoff:


labs/intro-to-ada/chapters/standard_library.rst
```diff
@@ -0,0 +1,1073 @@
...
+
+       Check (Test_Case_Index'Value (Argument (1)));
+    end Main;
+
+Inventory
+---------
+
+This exercise is based on the *Inventory* exercise from the
+:doc:`more_about_types`. Your goal is to rewrite that exercise using
+vectors |mdash| instead of arrays |mdash| and unbounded strings.
```
frank-at-adacore: Other than being more complicated than "to-do" list, why do we have another exercise for vectors? If we're just adding an example of "cursor" functionality, either one of these examples would work just as well with sets, and that gives the student more exposure to the different containers

labs/intro-to-ada/chapters/solutions/standard_library.rst
```diff
@@ -0,0 +1,1031 @@
...
+
+          return Total;
+       end Get;
+
+       function Get (Inv   : Inventory) return Amount
+       is
+          Total : Amount := 0.0;
+       begin
+          for C in Inv.List_Item.Iterate loop
+             Total :=  + Get (Inv, Item_ID (C));
```
frank-at-adacore: If someone looks at the solutions for help, do we want to add comments when things are confusing? I thought line 349 was an error, but it's just using OOP notation. But we don't do that on line 338. So maybe a comment indicating that?

labs/intro-to-ada/chapters/solutions/standard_library.rst
```diff
@@ -0,0 +1,1031 @@
...
+       begin
+          for C in Inv.List_Item.Iterate loop
+             declare
+                I : constant Item_ID := Item_ID (C);
+             begin
+                Put_Line ("==== ITEM "
+                          & ": " & Get (Inv, I));
+                New_Line;
+                Put ("Price:     ");
+                Put (Amount'(Get (Inv, I)), 1, 2, 0);
```
frank-at-adacore: This is wrong. There is no accessor function for the 'price' field - this is returning the 'Stock_Amount' value.
I made some mistakes, and the use of Amount and Quantity to mean different things was confusing (should use "value" instead of "amount", because for "Overall" saying "Amount Bought" usually means quantity, not cost)

labs/intro-to-ada/chapters/standard_library.rst
gusthoff: Yes, having a more complicated exercise was the main reason. However, I agree that we should have an exercise using sets, too. I'll work on that.

labs/intro-to-ada/chapters/solutions/standard_library.rst
gusthoff: Thanks for noticing this: it was a bug. I've just fixed it.

labs/intro-to-ada/chapters/solutions/standard_library.rst
gusthoff: Agreed, there's no _price_ field in that case. I've removed those lines. Also, I've replaced the wording ("amount"). Let me know what you think.

frank-at-adacore: Easier to understand now. thanks

PR #292, opened 2019-12-13
Widget: fixing HTML output

gusthoff:
Avoiding Latex code in case of HTML output

### DIFF
--- a/engine/sphinx/widget_extension.py
+++ b/engine/sphinx/widget_extension.py
@@ -279,28 +279,32 @@ class WidgetCodeDirective(Directive):
                 )
 
             nodes_latex = []
-            for f in files:
-                # Based on sphinx/directives/code.py
 
-                container_node = nodes.container(
-                    '', literal_block=True, classes=['literal-block-wrapper'])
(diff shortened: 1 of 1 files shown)


engine/sphinx/widget_extension.py
```diff
@@ -279,28 +279,32 @@ def get_shadow_div(basename, content):
                 )
 
             nodes_latex = []
-            for f in files:
-                # Based on sphinx/directives/code.py
 
-                container_node = nodes.container(
-                    '', literal_block=True, classes=['literal-block-wrapper'])
+            # Attemping to detect HTML or Latex output by checking for 'html' in tags
+            if 'html' not in self.state.state_machine.document.settings.env.app.tags.tags:
```
gusthoff: This looks ugly, but I haven't found a better solution.

Robert-Tice: Is there any information in the `options` dictionary attribute of the Directive? If not, then the sphinx creators maybe didn't plan for this sort of flexibility and we might be stuck with the workaround.

PR #297, opened 2020-01-10
Improving Latex/PDF output

gusthoff:


engine/sphinx/learn_theme/static/css/theme.css
```diff
@@ -49,6 +49,21 @@ a:hover, a:active {
     outline: 0
 }
 
+a.button {
```
Robert-Tice: This is a very generic css descriptor that has a potential to effect more than you intend it to. I would recommend changing the class from "button" to something much more specific to avoid ambiguity and accidentally effecting something that has the same signature that already exists.

PR #300, opened 2020-02-10
Adding simple example of variable declaration

gusthoff:


courses/intro-to-ada/chapters/imperative_language.rst
```diff
@@ -520,10 +540,9 @@ as declarative regions.
        --  Call to Nested
     end Main;
 
--  A declaration cannot appear as a statement if you
-   need to declare a local variable amidst the statements,
-   you can introduce a new declarative region with a block
-   statement:
+A declaration cannot appear as a statement if you need to declare a local
```
frank-at-adacore: Seems to be a run-on sentence. Should be "A declaration cannot appear as a statement. If you need to declare a local variable amidst the statements, you can introduce a new declarative region with a block statement:"

gusthoff: Interestingly, I didn't write this sentence, but just changed the formatting (it was rather a side-effect). :-) 

Nice catch! I'll correct this.

PR #301, opened 2020-02-11
Adding section on type aliases

gusthoff:


courses/intro-to-ada/chapters/strongly_typed_language.rst
```diff
@@ -658,3 +658,54 @@ an exception will be raised.
...
        --           Compiles, but exception at runtime
     end Greet;
+
+Type aliases
+------------
+
+Previously, we've seen that we can create new types by declaring
+:ada:`type Miles is new Float`. We could also create type aliases, which
+generate alternative names |mdash| *aliases* |mdash| for known types. We achieve
+this by using subtypes. In this case, however, we don't get the benefits of
```
frank-at-adacore: I would say "we don't get all of the benefits" - without "all of", implies (to me) no benefits - we still get the benefits from the parent

sttaft:
I'll admit to being somewhat uncomfortable with the term "type aliases."
Ada has subtypes and derived types.  Introducing yet another term seems
like it might add to the confusion.

-Tuck

gusthoff:
@sttaft : The term "type alias" is frequently used in other languages. Just to name a few examples:

- C: `typedef`
- C++: `using`
- Swift: `typealias`
- Kotlin: `typealias`
- Haskell: `type`

Since the _Introduction to Ada_ course assumes basic programming knowledge of other languages, reader that are familiar with the concept of type aliases --- based on their knowledge of other languages --- might wonder how this is done in Ada. In fact, this is rarely mentioned in online documentation (or books) about Ada, and it might not be obvious for most readers (it wasn't obvious to me when I started learning Ada). So I think it's important to show how this is done in Ada and make this explicit (by having a separate subsection).

I understand that you're reluctant to introduce this foreign concept to the Ada world. But, again, this is about creating a bridge between other languages and the Ada language.

sttaft:
I am just reluctant to introduce the term so visibly, when it really isn't
an Ada term.  I would rather see a title that is something like "Ada
Subtypes and Derived Types" and then inside say that these are similar to
what other languages call "type aliases."

-Tuck

sttaft:
To be more specific, I would suggest you change the title "Type Aliases"
to, for example, "Subtypes as Type Aliases" to make it clear that these are
Ada subtypes, that happen to impose no new constraint, and hence can act as
an alias.  Alternatively, you could make it a subsection of the Subtypes
section, to clarify that it is just a special case of subtypes.

-Tuck

gusthoff:
@sttaft : As you suggested, I integrated this section with the one on _subtypes_ **and** renamed the title to "Subtypes as Type Aliases." Also, I've added:

- a paragraph mentioning that this is only valid as long as we don't have new constraints;
- a small admonition mentioning how type aliases are declared in other languages.

sttaft:
Thanks, a lot, Gustavo.  The work you are doing is really great.  We have
been talking for years about creating such material, and it is wonderful to
see it really happening.

Take care,
-Tuck

gusthoff:
@sttaft : Thank you, too, for your suggestions: they helped me a lot!

PR #302, opened 2020-02-12
Including links to other sections

gusthoff:
Adding links to unbounded strings and dates for further information.


gusthoff:
Unfortunately, I had to solve a `git` conflct, which requires a new approval.

PR #313, opened 2020-04-01
Contact form

Robert-Tice:


quinot:
Before this PR is merged, can you clarify what the design is, from the standpoint of how data flows from users to us and then within AdaCore? We need to make sure that the processing of users' data is consistent with personal information regulations such as the GDPR, and with our privacy policy.

Robert-Tice:
Sure, will reply via email.

PR #326, opened 2020-04-17
Using GNAT switch instead of Assertion_Policy pragmas

gusthoff:


backend/app/widget/project.py
```diff
@@ -252,10 +252,10 @@ def build(self):
             Returns the status code returned from the build
         """
         rep = MQReporter(self.app, self.task_id)
-        rep.console(["gprbuild", "-q", "-P", self.gpr.get_name(), "-gnatwa"])
+        rep.console(["gprbuild", "-q", "-P", self.gpr.get_name(), "-gnatwa", "-gnata"])
 
         main_path = os.path.join(self.remote_tempd, self.gpr.get_name())
-        line = ["gprbuild", "-q", "-P", main_path, "-gnatwa"]
+        line = ["gprbuild", "-q", "-P", main_path, "-gnatwa", "-gnata"]
```
Robert-Tice, line 258: We don't want to add -gnata here. We should add it to the project file. We should probably add -gnatwa to the project file too to simplify this build line so that users can easily replicate similar behavior when they download the example. Can you move these two switches to the Compiler package in the project file: https://github.com/gusthoff/ada-learning-material/blob/intro_to_ada/review/contracts/20200417_pragmas/backend/app/widget/static/templates/inline_code/main.gpr

gusthoff: Thanks! I was actually looking for that file...

backend/app/widget/static/templates/inline_code/main.gpr
```diff
@@ -5,7 +5,7 @@ project Main is
    --LANGUAGE_PLACEHOLDER--
 
    package Compiler is
-      for Switches ("ada") use ("-g", "-O0", "-gnata");
+      for Default_Switches ("ada") use ("-g", "-O0", "-gnata");
```
gusthoff, line 8: "for Switches" is used for individual files.

Robert-Tice: If Switches is passed a language instead of a filename, it acts the same as Default_Switches.

PR #328, opened 2020-04-17
Modifying Ada main detection algorithm

Robert-Tice:
Its still hacky but now if the algorithm detects more than one Ada body without a corresponding package, it will look for one of the files to be called main.adb and it will use that one. If none are named main.adb, it will not have a main. If more than one is named main.adb it will throw an exception.

### DIFF
--- a/backend/app/widget/project.py
+++ b/backend/app/widget/project.py
@@ -131,12 +131,21 @@ class Project:
         mains = find_mains(self.file_list)
 
         if len(mains) > 1:
-            raise ProjectError("More than one main found in project")
+            main_match = [i for i in mains if "main.adb" in i]
+            if len(main_match) > 1:
+                raise ProjectError("More than one main.adb found in project")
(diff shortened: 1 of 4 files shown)


backend/app/widget/project.py
```diff
@@ -131,13 +131,22 @@ def __init__(self, files, spark_mode=False):
         mains = find_mains(self.file_list)
 
         if len(mains) > 1:
-            raise ProjectError("More than one main found in project")
+            main_match = [i for i in mains if "main.adb" in i]
+            if len(main_match) > 1:
+                raise ProjectError("More than one main.adb found in project")
```
gusthoff, line 136: When does this condition occur?

Robert-Tice: Ideally, never! But it could happen if there are two procedure Mains declared in an aggregate or lab setup. But it should never happen.

Issue #321, opened 2020-04-08
Error text highlight deselects on mouse depress


Robert-Tice:
Reported by: FC

When error text is highlighted in the output window and the mouse depresses on the error line, the text is automatically deselected. When the entire line is selected, and the mouse is depressed off the error line, the selection stays. This is happening because of the hooks that jump the editor to the source line displayed in the error text.

Robert-Tice:
Fixed in PR #332

PR #332, opened 2020-04-17
Fixing issue #321

Robert-Tice:
Fixing selection issue in output area. If text is selected, focus won't jump to the offending line in the editor. If nothing is selected, focus will jump as usual.


Robert-Tice:
This fixes #321

PR #335, opened 2020-04-20
Fixing excessive Put_Line noise crashing page rendering

Robert-Tice:
Making output_area scrollable with fixed max height.


Robert-Tice:
This possibly fixes it. We may need to test this and throttle check_output requests after threshold. The backend should interrupt the process before the browser blows through all its memory though.

PR #384, opened 2020-07-17
Adding section on the mixed-language code building & debugging

gusthoff:


content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -134,10 +134,40 @@ And that's all that's necessary. Here's an example of a call to :ada:`Call`:
 Building and Debugging mixed language code
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. todo::
+The easiest way to build an application using mixed C / Ada code is to create
+a simple project file for :program:`gprbuild` and specify C as an additional
+language. By default, when using :program:`gprbuild` we only compile Ada source
+files. To compile C code files as well, we use the ``Languages`` entry and
```
pat-rogers, line 140: I suggest "entry" should be "attribute" since "entry" is used elsewhere as a language construct and "attribute" is the actual term in case they look it up in the user guide.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -134,10 +134,40 @@ And that's all that's necessary. Here's an example of a call to :ada:`Call`:
...
+    end Default;
+
+Then, we use this project file to build the application by simply calling
+``gprbuild``. Alternatively, we can specify the project file on the
+command-line with the ``-P`` option |mdash| for example,
+``gprbuild -P default.gpr``. In both cases, ``gprbuild`` compiles all C
+source-code file found in the directory and links the corresponding object
+files to build the executable.
+
+In order to include debug information, you can use ``gprbuild -cargs -g``. This
```
pat-rogers, line 160: Perhaps use the Builder package for the "-g" switch?

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -134,10 +134,40 @@ And that's all that's necessary. Here's an example of a call to :ada:`Call`:
...
+option adds debug information based on both C and Ada code to the executable.
+To debug the executable, you can use applications that you would normally use
+to debug C-code-based executables, such as :program:`gdb` or :program:`ddd`. If
```
pat-rogers, line 163: Note that our GDB is enhanced for Ada (or at least it used to be, and likely still is) so there would be a difference from one the users gets with a C-only distribution.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
gusthoff: I've just included it as an alternative.

PR #386, opened 2020-07-24
Adding section on overlays for serialization

gusthoff:


content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -725,9 +725,467 @@ above could also be literally translated to:
...
+    with Ada.Text_IO; use Ada.Text_IO;
+
+    procedure Simple_Bitfield is
+       type Bit_Field is array (Natural range <>) of Boolean with Pack;
+
+       V : Integer := 0;
+       B : Bit_Field (0 .. V'Size - 1) with Address => V'Address;
+    begin
+       B (2) := True;
+       Put_Line ("V = " & Integer'Image (V));
```
QuentinOchem, line 780: For the record, this is a case where volatile on V may be needed, in particular at high level of optimizations.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -725,9 +725,467 @@ above could also be literally translated to:
...
+
+    !bitfield_serialization.c
+    #include <stdio.h>
+
+    #include "my_recs.h"
+    #include "serializer.h"
+
+    int main(int argc, const char * argv[])
+    {
+        rec r = {5, "abc"};
```
QuentinOchem, line 989: Should't the C string to be 4 characters len, to account for the \0 at the end? Otherwise it seems like we're writing one character to far...

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -725,9 +725,467 @@ above could also be literally translated to:
...
+                         R : out Rec) is
+          B_R : Bit_Field (0 .. R'Size - 1) with Address => R'Address;
+       begin
+          --  Assigning data from bit-field parameter B to bit-field
+          --  representation of output parameter.
+          B_R := B;
+       end To_Rec;
+
+       function To_Rec (B : Bit_Field) return Rec is
+          R   : Rec;
```
QuentinOchem, line 1075: Same as before, I would just give to R the address of B and return R directly. In this particular case, you're saving an extra copy :-) (which arguably would probably be optimized at high levels).

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -725,9 +725,467 @@ above could also be literally translated to:
...
+        to_r(&r1, sizeof(r1) * 8, &r2);
+
+        printf("New bitstream received!\n");
+        printf("r2 = ");
+        display_r (&r2);
+        printf("\n");
+    }
+
+Here, :c:`to_r` casts both pointer parameters to pointers to :c:`char` to get
+a byte-aligned pointer. Then, it simply copies the data byte-by-byte.
```
QuentinOchem, line 1191: I would suggest to add a section Overlays v.s. Unchecked_Conversions, explaining that Overlays are more efficient and allow to have one change automatically reflected by the other object, while Unchecked_Conversions require a copy. Unchecked_Conversions are not fundamentally bad and can also be used (they're actually cleaner and safer), but they're not always at the right level of performance / semantics. Perhaps that should be the introduction of this section.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -725,9 +725,467 @@ above could also be literally translated to:
...
+          begin
+             case V is
+                when False => Put ("0");
+                when True  => Put ("1");
+             end case;
+          end Show_Bit;
+
+       begin
+          Put ("Bits: ");
+          for I in B'Range loop
```
QuentinOchem, line 906: Perhaps it's a good place to use "for .. of " instead?

for I of B loop
  Show_Bit (I);
end loop;

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
gusthoff: You're right. I wanted to keep the C version as close as possible to the Ada version, but the way I'm initializing the string for `r` is still wrong (`strncpy` could be used instead to copy only 3 characters). In any case, a non null-terminated array of characters is not a valid C string, so I'll correct the example.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
gusthoff: Yes, that makes sense. I'll add this topic to a separate section and work on it in another PR. Also, I'll add a link between the sections.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -762,19 +762,53 @@ This technique is called overlays for serialization. Now, any operation that we
...
+initialization which could overwrite the existing content while creating the
+overlay |mdash| see an example in the admonition below. The declaration now
+becomes:
+
+.. code-block:: ada
+
+    B : Bit_Field (0 .. V'Size - 1)
+      with
+        Address => V'Address,
+        Import  => True,
```
QuentinOchem, line 795: Style note, I don't think you need Import => True. Import by itself implies True.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -816,20 +851,138 @@ In C, we would rely on bit-shifting and masking to set that specific bit:
         printf("v = %d\n", v);
     }
 
+.. admonition:: Important
+
+    The :ada:`Default_Value` aspect is associated with type declarations. When
```
QuentinOchem, line 856: I would start at a higher level, something like:

Ada has the concept of default initialization, which provides default values for either a type of its components. This is the case for example when specifying default values for components of record:

type X is record
   v : Integer := 0;
end record;

when associating a type with a Default_Value aspect, or even by default on certain types (e.g. access types have a default value of null).

PR #411, opened 2020-09-16
Corrected "T.Put_Line" to "TIO.Put_Line"

jstuyts:
Text below the example used `T.Put_Line`, but the example renames to `TIO`.


gusthoff (approved):
Thanks for the correction!

PR #414, opened 2020-09-16
Changed "parameters mode" to singular to match rest of documentation

jstuyts:
The documentation talks about _parameter modes_, but the section discussing them used _parameters modes_. I changed the latter.


gusthoff (approved):
Thanks for this correction as well!

PR #424, opened 2020-09-28
Ace gutter annotations

Robert-Tice:
This uses ace editor's gutter annotations feature to add gutter annotations based on server feedback. Info messages will add a little "i" symbol next to lines with info messages (see SPARK courses for examples). Warnings will add a little yellow triangle. Errors will add a red x. All icons have mouse over text which indicates the error message reported by the server.


pat-rogers (approved):
cool

setton:
very nice

PR #429, opened 2020-10-02
Adding section on switches and optimizations

gusthoff:


gusthoff:
Let me know if there are other switches I might be missing here.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/08_Performance.rst
```diff
@@ -29,10 +29,103 @@ suspects and their resolution.
...
+
+In addition to this aspect, we need to set the optimization level to at least
+``-O1`` and use the ``-gnatn`` switch, which instructs the compiler to take the
+:ada:`Inline` aspect into account.
+
+Note, however, that the :ada:`Inline` aspect is just a *recommendation* to the
+compiler. Sometimes, the compiler might not be able to follow this
+recommendation, so it won't inline the subprogram. If we want to ensure that a
+subprogram is always inlined or else get a compilation error when it isn't, we
+need to use the :ada:`Inline_Always` aspect instead of :ada:`Inline`.
```
pat-rogers, line 128: I was told (by Robert, who wrote the compiler) not to use Inline_Always for that reason because it was intended only for cases in which the functionality would be incorrect absent the inlining. So use it on procedures that insert assembly instructions that only make sense inline, such as memory barriers, but nothing else.
With the aspect/pragma Inline and and the necessary switches we do get a warning when inline cannot be achieved.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/08_Performance.rst
```diff
@@ -29,10 +29,103 @@ suspects and their resolution.
...
+In addition to the levels presented above, GNAT also has the ``-Os`` switch,
+which allows for optimizing code and data usage.
+
+Inlining
+~~~~~~~~
+
+As we've seen in the previous section, the highest optimization level (``-O3``)
+tries to automatic inline subprograms whenever possible. There might be
+situations, however, where this doesn't happen automatically. In those cases,
+manual inlining may help.
```
pat-rogers, line 75: The problem with O3 is that it inlines too much rather than not enough.  As a result, the cache becomes an issue and the code can slow down as a result.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/08_Performance.rst
```diff
@@ -29,10 +29,103 @@ suspects and their resolution.
...
+When compiling this example, the GNAT will inline :ada:`Average` in the
+:ada:`Compute_Average` procedure.
+
+In addition to this aspect, we need to set the optimization level to at least
+``-O1`` and use the ``-gnatn`` switch, which instructs the compiler to take the
+:ada:`Inline` aspect into account.
+
+Note, however, that the :ada:`Inline` aspect is just a *recommendation* to the
+compiler. Sometimes, the compiler might not be able to follow this
+recommendation, so it won't inline the subprogram. If we want to ensure that a
```
pat-rogers, line 126: Maybe say why the compiler might not: the code is too large, too complicated (eg exception handling), contains tasks, etc.

gusthoff:
Thanks a lot for all the hints, I tried to incorporate them. I certainly learned a few new things today.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/08_Performance.rst
```diff
@@ -141,6 +141,17 @@ inline a subprogram:
 
 - when it contains tasks, etc.
 
+In addition to the :ada:`Inline` aspect, we also have the :ada:`Inline_Always`
```
pat-rogers, line 144: In retrospect maybe we shouldn't even mention Inline_Always. It's a fairly advanced topic.

pat-rogers (approved):

PR #435, opened 2020-10-30
Adding section on overlays vs. unchecked conversions

gusthoff:


content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -1395,16 +1395,145 @@ a byte-aligned pointer. Then, it simply copies the data byte-by-byte.
...
+conversion is happening, since the conversion is written explicitly in the
+code. With overlays, that conversion is automatic and therefore implicit. In
+that sense, using :ada:`Unchecked_Conversion` is a cleaner and safer approach.
+On the other hand, :ada:`Unchecked_Conversion` requires a copy, so it's less
+efficient than overlays, where no copy is performed |mdash| because one change
+in the source object is automatically reflected in the target object (and
+vice-versa). In the end, the choice between unchecked conversions and overlays
+depends on the level of performance that you want to achieve.
+
+Also note that :ada:`Unchecked_Conversion` only works with scalar types.
```
pat-rogers, line 1535: Why does this say that it only works with scalar types?

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -1395,16 +1395,145 @@ a byte-aligned pointer. Then, it simply copies the data byte-by-byte.
...
+    with Ada.Text_IO;              use Ada.Text_IO;
+    with Ada.Unchecked_Conversion;
+
+    procedure Simple_Unchecked_Conversion is
+       type State is (Off, State_1, State_2)
+         with Size => Integer'Size;
+
+       for State use (Off => 0, State_1 => 32, State_2 => 64);
+
+       function To_Integer is new Ada.Unchecked_Conversion (Source => State,
```
pat-rogers, line 1418: Just an aside. Consider using a naming scheme of "As_*" instead of "To_*" because we aren't really converting the value (unless we then assign the result of the conversion to something else). Instead, we are asking the compiler to treat the value as if it is of a different type, so that we can manipulate/query it in a manner consistent with that other type. For example:

type Int_Pointer is access all Integer;
function As_Int_Pointer is new Ada.Unchecked_Conversion (Source => System.Address, Target => Int_Pointer);

Then, given an address X, hopefully of some integer, we can say As_Int_Pointer (X).all ;= 42; -- the answer to life, the universe, ...

PR #437, opened 2020-10-30
GitHub actions

Robert-Tice:
Replacing travis tests with GitHub actions equivalents. Adding back the sphinx content tests which are currently failing.


gusthoff (commented):
In general, I'm ok with the structure. We need to make sure, however, that GNAT Community 2020 is used here, since all tests for the code example are passing when using that version of the compiler.

gusthoff (approved):
Perfect, thanks a lot!

PR #439, opened 2020-11-06
Adding section on volatile and atomic data

gusthoff:


content/courses/Ada_For_The_C_Embedded_Developer/chapters/04_Embedded.rst
```diff
@@ -240,9 +240,194 @@ Dealing with Absence of FPU with Fixed Point
...
+example is the one of
+`memory-mapped I/O <https://en.wikipedia.org/wiki/Memory-mapped_I/O>`_, where
+the hardware might be constantly changing the value of an object in memory.
+
+Because the value of a volatile object may be constantly changing, a compiler
+cannot store the value of that object into a register and use the value from
+the register in subsequent operations |mdash| because the value stored in the
+register would be outdated if another process had changed the volatile object
+in the meantime. Instead, the compiler must read the value of that object from
+memory for each access. In Ada, the standard specifies that, "for a volatile
```
pat-rogers, line 264: Unfortunately (or fortunately, depending on who you ask), the Ada RM no longer says that. The last version that did was Ada 2005. The wording is a good bit more subtle now, partly for the sake of multiprocessors. This is a very complicated topic...

content/courses/Ada_For_The_C_Embedded_Developer/chapters/04_Embedded.rst
```diff
@@ -240,9 +240,194 @@ Dealing with Absence of FPU with Fixed Point
...
+
+.. code-block:: ada
+
+    Arr : array (1 .. 2) of Long_Float with Volatile;
+
+Atomic
+~~~~~~
+
+Operations are said to be atomic when they can be completed without
+interruptions. This is an important requirement when we're performing
```
pat-rogers, line 378: Regarding operations, that's a different use of the term "atomic" (eg, for a database update operation) than what the RM is talking about.  Here, atomic means that any read or update of such an object must be done in the generated assembly language such that no other instruction could execute on that same object before the read or update completes. (That's not the right wording, but that's the idea.) So generally that means read or update with one assembly language instruction. For example, if we have a 32-bit object composed of four 8-bit bytes, the compiler cannot generate code to read or update the object using four 8-bit store/load instructions, or even two 16-bit store/load instructions. 
That means that there are objects for which the Atomic aspect cannot be specified. Lots of machines support integer types that are larger than the native word-sized integer. For example, a 16-bit machine probably supports both 16-bit and 32-bit integers, but only 16-bit integer objects can be marked as atomic. And more generally, only objects that fit into at most 16-bits, on such a machine.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/04_Embedded.rst
```diff
@@ -240,9 +240,194 @@ Dealing with Absence of FPU with Fixed Point
...
+operations on objects in memory that are shared between multiple processes.
+An atomic object is an object that only accepts atomic operations. The Ada
```
pat-rogers, line 380: "operations" is too broad, it really means reads and updates only

content/courses/Ada_For_The_C_Embedded_Developer/chapters/04_Embedded.rst
```diff
@@ -240,9 +240,194 @@ Dealing with Absence of FPU with Fixed Point
...
+standard specifies that "for an atomic object (including an atomic component),
+all reads and updates of the object as a whole are indivisible."
+
+Atomicity is important, for example, when dealing with shared hardware
+registers. In Ada, we can use the :ada:`Atomic` aspect to indicate that an
+object is atomic. This is how we can use the aspect to declare a shared
```
pat-rogers, line 386: Yes, but not in all cases. Only when the hardware requires it, in other words. Atomicity isn't an intrinsic property of memory-mapped registers. There certainly are cases, and some are in the STM32 register set, but that's not always the case. The SVD code generator just assumes the worst and makes them all as such.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/04_Embedded.rst
gusthoff: Thanks for this explanation; it is very good! Actually, my understanding was the same as yours, but I wasn't using the correct terminology.

I decided to use your explanation in an admonition, so the context becomes clearer to the readers.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/04_Embedded.rst
gusthoff: Thanks! I adapted the paragraph accordingly.

pat-rogers (approved):

PR #467, opened 2020-12-15
Ada for the C Embedded Developer course review

filip-gajowniczek-adacore:
General comments forthcoming.


backend/dev_server.sh
```diff
@@ -7,3 +7,4 @@ flask run --host=0.0.0.0 &
 celery worker -A celery_worker.celery -E --loglevel=DEBUG &
 
 wait
+6
```
Robert-Tice, line 10: This looks like a typo?

filip-gajowniczek-adacore: yessir

content/courses/Ada_For_The_C_Embedded_Developer/chapters/02_Perspective.rst
```diff
@@ -148,17 +148,17 @@ The first line of the Ada code is giving us access to the :ada:`Ada.Text_IO`
...
 
 You may have noticed that the Ada syntax is more verbose than C. Instead of
 using braces :c:`{}` to declare scope, Ada uses keywords. :ada:`is` opens a
 declarative scope |mdash| which is empty here as there's no variable to
 declare. :ada:`begin` opens a sequence of statements. Within this sequence,
-we're calling the function :ada:`Put_Line`, prefixing explicitly by the name of
+we're calling the function :ada:`Put_Line`, prefixing explicitly with the name of
 the library unit where it's declared, :ada:`Ada.Text_IO`. The absence of the
-end of line :c:`\n` can also be noted, as :ada:`Put_Line` always terminates by
+end of line ``\n`` can also be noted, as :ada:`Put_Line` always terminates by
```
filip-gajowniczek-adacore, line 161: Please not the change in syntax to render the newline. This will render it in a different "style" compared to the other inlined text which uses the :c:`` syntax.

filip-gajowniczek-adacore:
Ada aspects are mentioned throughout the course, but never formally introduced. What does the team think of adding a section explaining what an Ada aspect is, perhaps in chapter 2?

filip-gajowniczek-adacore:
In the Arrays and Strings section. The last example initializes an unconstrained array with 9 elements. It may be interesting to draw the readers attention to the fact that the indexing values are arbitrary.

filip-gajowniczek-adacore:
The section on pointers concludes with examples of manually deallocating memory in Ada with one of the generic packages. It mentions generics without having introduced them formally as a concept. A reference to the generic section may be appropriate here, or reorganization of the content.

filip-gajowniczek-adacore:
Suggest that examples under the "Dealing with Absence of FPU with Fixed Point" section should either be made to run or turned into code-blocks and have the text revised.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/02_Perspective.rst
gusthoff: In general, we use language-specific syntax highlighting for any piece of code that might refer to source-code. For example, if we're referring to variable X in a previous code example, we'd write:
```
Variable :ada:`X` in the example above is [...]
```
We could apply the same principle here for `\n`. That being said, however, I don't have a strong opinion here. :-)

content/courses/Ada_For_The_C_Embedded_Developer/chapters/05_SPARK.rst
```diff
@@ -946,18 +946,6 @@ Now we don't need the defensive code in the procedure body. That's safe
...
-of software engineering apply) we would take advantage of the
-compile-time visibility controls that packages offer. Specifically, we
-would move all the variables' declarations to the private part of the
-package, or even the package body, so that client code could not
-possibly access the array directly. Only procedure :ada:`Read` would
-remain visible to clients, thus remaining the only means of accessing
-the array. However, that change would entail others, and in this chapter
-we are only concerned with introducing the capabilities of SPARK.
-Therefore, we keep the examples as simple as possible.
-
```
gusthoff, line 960: I'd like to understand why you're removing this paragraph.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -417,8 +417,8 @@ names in their type.
 In Ada, the package is a namespace |mdash| two entities declared in two
 different packages are clearly identified and can always be specifically
 designated. The C names are usually a good indication of the names of the
-futures packages and should be stripped |mdash| it will be possible to force
-full name if useful. For example, here's how the following declaration and call
+futures packages and should be stripped |mdash| it is possible to the full name
```
gusthoff, line 420: It should probably be:

"it is possible to **force** the full name" 

instead.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -622,8 +622,8 @@ pieces of data into a larger structure. In Ada, this can be done by describing
 the structure layout at the type level through representation clauses, and then
 accessing this structure as any other.
 
-This is the case of an array of bytes representing various flags. In C, this
-would be done through masks, e.g.:
+Consider the case of using a C primitive type's bits to represent various
```
gusthoff, line 625: I agree the original phrase isn't good. However, this part:

> using a C primitive type's bits

is a little bit confusing.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
```diff
@@ -1167,7 +1167,7 @@ shifting and masking to access the bits of that byte. Here, we use the
 
     void transmit (void *bits, int len)
     {
-        int i, j;
+        int i;
```
gusthoff, line 1170: We're actually using ``j`` below...

gusthoff (commented):
Thanks a lot for all the corrections!

I've added a couple of editorial changes (primarily of editorial nature). Also, I've got a couple of questions (see inline conversations).

content/courses/Ada_For_The_C_Embedded_Developer/chapters/02_Perspective.rst
filip-gajowniczek-adacore: Agree that this is the desired approach! However, when I tried using the :ada: and :c: syntax highlighting there seemed to be an issue escaping the ```\``` character.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/05_SPARK.rst
filip-gajowniczek-adacore: In my opinion the user is reading this section to learn about SPARK, not necessarily about software design principles.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
filip-gajowniczek-adacore: https://en.wikipedia.org/wiki/Bit_field#C_programming_language
https://en.cppreference.com/w/cpp/language/bit_field

The type defines the size of the memory into which the bitfields will be mapped (and implementation specific of how those bitfields are interpreted as signed or unsigned).

Perhaps it could be reworded to something like:
" Consider the case of using a C primitive type as a container for single bit boolean flags"?

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
filip-gajowniczek-adacore: Investigating :) I think I may have made this change to the wrong code block.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
filip-gajowniczek-adacore: Is there actually a mechanism to force such adoption? Or is it just that the usage of the full name is permitted? I definitely made a typo here though that needs to be fixed.

content/courses/Ada_For_The_C_Embedded_Developer/chapters/02_Perspective.rst
```diff
@@ -2310,7 +2337,7 @@ Pointers to scalar objects in Ada and C look like:
 
 [Ada]
 
-.. code:: ada run_button project=Courses.Ada_For_C_Embedded_Dev.Perspective.Dereferencing_Ada
+.. code-block:: ada
```
filip-gajowniczek-adacore, line 2340: revert the code blocks back to code and make everything runnable

content/courses/Ada_For_The_C_Embedded_Developer/chapters/05_SPARK.rst
filip-gajowniczek-adacore: reinstate

content/courses/Ada_For_The_C_Embedded_Developer/chapters/02_Perspective.rst
gusthoff: In that case, let's just use your change. :-)

content/courses/Ada_For_The_C_Embedded_Developer/chapters/06_Translation.rst
gusthoff: Yes, the intention was actually clear. ;-) Just that specific expression ("a C primitive type's bits") was confusing. But I do like your rewording; you could use that.

gusthoff (approved):
Thanks again for your thorough review!

PR #493, opened 2021-01-15
Sphinx: adding training examples to the "exclude" list.

gusthoff:


gusthoff:
Cancelling because we do want to have those examples on the production site.

PR #500, opened 2021-02-02
Intro to ada lab review

filip-gajowniczek-adacore:
General comments:

- Can we make code that isn't meant to be changed read-only.
- When comparing the user's output and the expected output, can we use a diff that highlights errors (easier to find whitespace errors, etc.)
- Would adding links to relevant resources applicable to the lab improve usability?
- Can we add a dialog box that prevents the user from accidentally navigating away from the page and losing their implementation?
- Can we change the embedded editors auto-complete functionality? Specifically when using the tick operator it tends to add a second tick which might frustrate the user.
- When the test output is displayed after submission, newlines are shown as spaces. This makes it difficult to determine white space errors in several exercises.
- The content should be reviewed for a consistent format in defining the lab exercises (Goal, Requirements, Remarks, Steps)
- When should we add ```with``` clauses for the users, and when should the user add them? Seems like it is inconsistent.
- Can the back-end be made more robust, so that it limits the output text if a user decides to add logging to debug their implementation? Otherwise the page may try to display immense amounts of output which causes shenanigans.
- Throughout the labs there are a lot of examples with "lists", but these aren't linked lists. Might this confuse the user? Should the term "collection" be used instead?

### DIFF
--- a/content/labs/intro-to-ada/chapters/arrays.rst
+++ b/content/labs/intro-to-ada/chapters/arrays.rst
@@ -810,6 +810,10 @@ String_10
 
     #. Function :ada:`To_String_10` returns constrained strings of
        :ada:`String_10` type based on an input parameter of :ada:`String` type.
+	   
+	   - For strings that are more than 10 characters, omit everything after the 11th character.
+	   
+	   - For strings that are fewer than 10 characters, pad the string with ' ' characters until it is 10 characters.
 
 **Remarks**:
 
--- a/content/labs/intro-to-ada/chapters/arrays.rst
+++ b/content/labs/intro-to-ada/chapters/arrays.rst
@@ -811,9 +811,11 @@ String_10
     #. Function :ada:`To_String_10` returns constrained strings of
        :ada:`String_10` type based on an input parameter of :ada:`String` type.
 	   
-	   - For strings that are more than 10 characters, omit everything after the 11th character.
+	   - For strings that are more than 10 characters, omit everything 
+	     after the 11th character.
 	   
-	   - For strings that are fewer than 10 characters, pad the string with ' ' characters until it is 10 characters.
+	   - For strings that are fewer than 10 characters, pad the string 
+	     with ' ' characters until it is 10 characters.
 
 **Remarks**:
 
@@ -1122,4 +1124,3 @@ List of Names
 
        Check (Test_Case_Index'Value (Argument (1)));
     end Main;
-
--- a/content/labs/intro-to-ada/chapters/records.rst
+++ b/content/labs/intro-to-ada/chapters/records.rst
@@ -414,7 +414,8 @@ keep track of assets.
     #. Procedure :ada:`Add` adds an item to the assets.
 
         #. Since we want to keep track of the assets, the implementation must
-           accumulate the total value of each item's inventory, the result of multiplying the item quantity and its price.
+           accumulate the total value of each item's inventory, the result of 
+           multiplying the item quantity and its price.
 
 .. code:: ada lab=Records.Inventory
 
--- a/content/labs/intro-to-ada/chapters/records.rst
+++ b/content/labs/intro-to-ada/chapters/records.rst
@@ -36,8 +36,7 @@ Directions
 **Remarks**:
 
     #. We make use of the algorithm implemented in the :ada:`Check_Direction`
-       procedure 
-	   (:doc:`chapter on imperative language <courses/intro-to-ada/chapters/imperative_language>`).
+       procedure (:doc:`chapter on imperative language <courses/intro-to-ada/chapters/imperative_language>`).
 
     #. For the sake of this exercise, we use the concept of *extended angles*.
        This includes the actual geometric angle and the corresponding direction
--- a/content/labs/intro-to-ada/chapters/records.rst
+++ b/content/labs/intro-to-ada/chapters/records.rst
@@ -36,8 +36,8 @@ Directions
 **Remarks**:
 
     #. We make use of the algorithm implemented in the :ada:`Check_Direction`
-       procedure (see chapter on
-       :doc:`chapter on imperative language <courses/intro-to-ada/chapters/imperative_language>`).
+       procedure 
+	   (:doc:`chapter on imperative language <courses/intro-to-ada/chapters/imperative_language>`).
 
     #. For the sake of this exercise, we use the concept of *extended angles*.
        This includes the actual geometric angle and the corresponding direction
@@ -415,7 +415,7 @@ keep track of assets.
     #. Procedure :ada:`Add` adds an item to the assets.
 
         #. Since we want to keep track of the assets, the implementation must
-           accumulate the total amount of each item.
+           accumulate the total value of each item's inventory, the product of the item quanitity and its price.
 
 .. code:: ada lab=Records.Inventory
 
--- a/content/labs/intro-to-ada/chapters/records.rst
+++ b/content/labs/intro-to-ada/chapters/records.rst
@@ -414,7 +414,7 @@ keep track of assets.
     #. Procedure :ada:`Add` adds an item to the assets.
 
         #. Since we want to keep track of the assets, the implementation must
-           accumulate the total value of each item's inventory, the product of the item quanitity and its price.
+           accumulate the total value of each item's inventory, the result of multiplying the item quantity and its price.
 
 .. code:: ada lab=Records.Inventory
 
--- a/content/labs/intro-to-ada/chapters/standard_library_containers.rst
+++ b/content/labs/intro-to-ada/chapters/standard_library_containers.rst
@@ -130,7 +130,8 @@ Simple todo list
 List of unique integers
 -----------------------
 
-**Goal**: create function that removes duplicates from and orders a collection of elements.
+**Goal**: create function that removes duplicates from and orders a collection 
+of elements.
 
 **Steps**:
 
@@ -272,4 +273,3 @@ List of unique integers
           end;
        end if;
     end Main;
-
--- a/content/labs/intro-to-ada/chapters/standard_library_containers.rst
+++ b/content/labs/intro-to-ada/chapters/standard_library_containers.rst
@@ -130,7 +130,7 @@ Simple todo list
 List of unique integers
 -----------------------
 
-**Goal**: create function that removes duplicates from and order elements.
+**Goal**: create function that removes duplicates from and orders a collection of elements.
 
 **Steps**:
 
--- a/content/labs/intro-to-ada/chapters/standard_library_numerics.rst
+++ b/content/labs/intro-to-ada/chapters/standard_library_numerics.rst
@@ -401,7 +401,9 @@ circle after rotation.
                 Point #4:  -90.00 degrees (= 270 degrees)
                 Point #5:    0.00 degrees
 
-		#. To rotate a complex number simply multiply it by a unit vector whose arg is the radian angle to be rotated: :math:`Z = e^\frac{2 \pi}{N}`
+            #. To rotate a complex number simply multiply it by a unit vector
+               whose arg is the radian angle to be rotated:
+               :math:`Z = e^\frac{2 \pi}{N}`
 
 .. code:: ada lab=Standard_Library.Rotation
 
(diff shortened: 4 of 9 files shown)


filip-gajowniczek-adacore:
States #3
- Difficult to tell where the user should insert code

filip-gajowniczek-adacore:
Operations

- Add the goals for this exercise
- The output format is not specified in requirements and can be frustrating to match correctly.

filip-gajowniczek-adacore:
Colors

- Add a requirement that the Colors need to be declared in the order provided.

filip-gajowniczek-adacore:
Temperatures

- The requirement on the Celsius_Int type seem odd, I interpreted it as rounding down for all values, but rounding up for values in (5504.5, 5504.85]

filip-gajowniczek-adacore:
Product Info

- Function signatures are already provided in the spec, so why not also provide them in the body to save the user copy and pasting?
- The function that takes paramter Product_Infos and out Currency_Array makes an assumption that the arrays are indexed over the same values, this should be made explicit.

filip-gajowniczek-adacore:
List of Names

- Parts of this exercise dealing with the Name_Type seem redundant with the previous String_10 exercise (i.e. conversion from String to fixed-length string)

filip-gajowniczek-adacore:
Aggregate Initialization

- Consider pre-supplying the function specifciations and bodies. What is the value added of having the user do this in this exercise?

filip-gajowniczek-adacore:
Simple todo list

- Add a hint about the initialized allocation syntax
- The requirements for Display procedure are unclear, might lead to frustration when users doesn't add "TO-DO LIST"

filip-gajowniczek-adacore:
Price List

- This example is redundant with previous ones where the user is asked to implement list management; I think the goal of this exercise is to highlight how discriminated records can have dynamic structures, and this can be done by just having the user implement the Get function.
- A candidate for addding the ```with Ada.Text_IO``` clause so the user doesn't have to remember to do that.

filip-gajowniczek-adacore:
Directions

- An explicit example of the target of the remark should be provided in the solution.

filip-gajowniczek-adacore:
Limited Strings

- Remark 3.4.1 doesn't mean much to a beginner and should be supplemented with an example.
- I believe this exercise and its solution assume that the string parameters of the Copy procedure have the same start index. This should be made explicit.

filip-gajowniczek-adacore:
Privacy > Bonus Exercise

- Consider adding code blocks here for the user's ease of use.

filip-gajowniczek-adacore:
Display Array

- This may be due to inexperience, but I do not like the pattern where a procedure is genericized by an index type, an element type, and then an array type when the array type is "deducible" from the previous two formal types. I would recommend redesigning this example, maybe by using an array as the generic formal type or implementing something like a swap procedure.

filip-gajowniczek-adacore:
Generic List
- The unconstrainted array formal type is missing from Appendix D of the Introduction To Ada course.
- As before, I am not a fan of the design pattern of genericizing on the array type and also having the client code provide the array as a an object formal parameter. Why not encapsulate the array and its management within the package?

filip-gajowniczek-adacore:
Eceptions > Unitialized Value

- Add a requirement for the message format the exception should be raised with
- Candidate for adding ```with Ada.Text_IO``` so the user doesn't have to.

filip-gajowniczek-adacore:
Re-raising Exceptions

- Seemed odd to me that the declaration of Another_Exception was in the spec of a package that never uses it. Would recommend reworking so that the exception is declared in check_exception.ads
- The requirements need to be fixed. Requirement 2.1 does not apply to the Constraint_Error case based on the solution's expected output.

filip-gajowniczek-adacore:
Display Service

- Candidate for adding ```with Ada.Text_IO;``` so that the user doesn't have to.
- Would add requirement that the service should support being entered multiple times.

filip-gajowniczek-adacore:
Event Manager

- Candidate for adding ```with Ada.Text_IO;``` so that the user doesn't have to.
- I wasn't able to understand why adding a delay in the Event entry "handler" caused my solution to fail

filip-gajowniczek-adacore:
Generic Protected Queue

- I would specify the queueing policy FIFO vs LIFO just in case.
- I would also modify the example to be a LIFO stack because it is easier to implement. And the complexities of FIFO implementation can avoid detracting the user from learning about tasking.

filip-gajowniczek-adacore:
Primary Color
- Remove requirements that are already satisfied by the code that should be changed (more suitable as remarks?) The user can get intimidated and use more mental effort to determine what they actually have to do.

filip-gajowniczek-adacore:
Simple Type Extension

- The solution provided might not actually meet the requirement of "keeping the values in sync" since the client code has direct access to the tagged types elements.
- The initialization example highlights the absence of destructor/constructor in Ada. If thats the goal, keep it. If thats not the goal, then I recommend foregoing the initialization routines and just having the display procedure and no "synchronization" requirement.

filip-gajowniczek-adacore:
Online Store

- I think the goal here is to teach the user about the dot notation usable with tagged types? I think a better way than having them implement a test, is for the test to use an "Image/Display" function whose paramter is the tagged type. The way it is right now, the user could just generate the output from the test inputs without actually instantiating any tagged types.

filip-gajowniczek-adacore:
Containers > Simple to do list

- Should this exercises "steps" be renamed to "requirements"?

filip-gajowniczek-adacore:
List of Events

- Inconsistency with which packages are ```with```ed for the user. Calendar is but Ordered_Map isn't.
- This example seems more appropriate in the containers section, as an example teaching about ordered maps.

filip-gajowniczek-adacore:
Rotation

- Requirements are provided for types the user is not expected to change. Recommend moving them to remarks.

content/labs/intro-to-ada/chapters/standard_library_numerics.rst
```diff
@@ -401,6 +401,8 @@ circle after rotation.
                 Point #4:  -90.00 degrees (= 270 degrees)
                 Point #5:    0.00 degrees
 
+		#. To rotate a complex number simply multiply it by a unit vector whose arg is the radian angle to be rotated: :math:`Z = e^\frac{2 \pi}{N}`
```
filip-gajowniczek-adacore, line 404: This doesn't display properly, and I couldn't figure out why. Looking for some help :)

content/labs/intro-to-ada/chapters/arrays.rst
```diff
@@ -810,6 +810,10 @@ String_10
 
     #. Function :ada:`To_String_10` returns constrained strings of
        :ada:`String_10` type based on an input parameter of :ada:`String` type.
+	   
+	   - For strings that are more than 10 characters, omit everything after the 11th character.
+	   
+	   - For strings that are fewer than 10 characters, pad the string with ' ' characters until it is 10 characters.
```
gusthoff, line 816:
```suggestion
	   
	   - For strings that are more than 10 characters, omit everything 
	     after the 11th character.
	   
	   - For strings that are fewer than 10 characters, pad the string 
	     with ' ' characters until it is 10 characters.
```

gusthoff: Please be aware of the typical 80-character limitation.

content/labs/intro-to-ada/chapters/records.rst
```diff
@@ -415,7 +414,7 @@ keep track of assets.
     #. Procedure :ada:`Add` adds an item to the assets.
 
         #. Since we want to keep track of the assets, the implementation must
-           accumulate the total amount of each item.
+           accumulate the total value of each item's inventory, the product of the item quanitity and its price.
```
gusthoff, line 417:
```suggestion
           accumulate the total value of each item's inventory, which is the 
           multiplication of the item quantity and its price.
```
Minor suggestion: I'd replace "product" by "multiplication" in this case. Also, there was a typo in "quantity."

content/labs/intro-to-ada/chapters/standard_library_containers.rst
```diff
@@ -130,7 +130,7 @@ Simple todo list
 List of unique integers
 -----------------------
 
-**Goal**: create function that removes duplicates from and order elements.
+**Goal**: create function that removes duplicates from and orders a collection of elements.
```
gusthoff, line 133:
```suggestion
**Goal**: create function that removes duplicates from and orders a collection 
of elements.
```

content/labs/intro-to-ada/chapters/records.rst
filip-gajowniczek-adacore: I agree that "product" could be confused with "item" in this domain. 

What do you think of "result of multiplying the item quantity and its price".

gusthoff: This sounds good.

filip-gajowniczek-adacore, line 417:
```suggestion
           accumulate the total value of each item's inventory, the result of multiplying the item quantity and its price.
```

content/labs/intro-to-ada/chapters/standard_library_numerics.rst
gusthoff:
```suggestion
            #. To rotate a complex number simply multiply it by a unit vector
               whose arg is the radian angle to be rotated:
               :math:`Z = e^\frac{2 \pi}{N}`
```

gusthoff: The paragraph with the formula was aligned with the previous code block...

gusthoff: In any case, I'm not totally convinced that the formula is actually useful here. Isn't the previous explanation (in simpler terms) enough?

content/labs/intro-to-ada/chapters/tasking.rst
```diff
@@ -230,7 +230,7 @@ Generic Protected Queue
 
         #. Entry :ada:`Push` stores an element in the queue.
 
-        #. Entry :ada:`Pop` removes an element from the queue.
+        #. Entry :ada:`Pop` removes an element from the queue and returns the element via output parameter.
```
gusthoff, line 233:
```suggestion
        #. Entry :ada:`Pop` removes an element from the queue and returns the 
           element via output parameter.
```

content/labs/intro-to-ada/chapters/records.rst
```diff
@@ -415,7 +414,7 @@ keep track of assets.
     #. Procedure :ada:`Add` adds an item to the assets.
 
         #. Since we want to keep track of the assets, the implementation must
-           accumulate the total amount of each item.
+           accumulate the total value of each item's inventory, the result of multiplying the item quantity and its price.
```
gusthoff, line 417:
```suggestion
           accumulate the total value of each item's inventory, the result of 
           multiplying the item quantity and its price.
```

gusthoff: Just adapting to the 80-character formatting... ;-)

content/labs/intro-to-ada/chapters/standard_library_numerics.rst
filip-gajowniczek-adacore: I think it depends on the user's knowledge of complex numbers; if we assume the reader remembers complex number operations then this would be unnecessary.

But the fact that we are describing the relationship between the coordinate form and the polar form suggests we assume no experience with complex numbers. 

In this case the reader understands that they want to rotate the vector through those angles, but they might not know how to do it. So thats what the 3rd point is meant to give them.

content/labs/intro-to-ada/chapters/records.rst
filip-gajowniczek-adacore: ahhh I'm sorry, I need to get better at that!

content/labs/intro-to-ada/chapters/standard_library_numerics.rst
gusthoff: Ok, fine with me. :-)

gusthoff:
I think we're ready to approve this PR. I'll take your comments above into consideration separately.

gusthoff (approved):

PR #502, opened 2021-02-05
Introducing support for testing code examples with "prove" buttons.

gusthoff:
- Adding Sphinx class: expected prove error.
- Adding support for compiler switches.


frontend/tests/compile_blocks.py
```diff
@@ -203,6 +219,74 @@ def __repr__(self):
 
 args.rst_files = [os.path.abspath(f) for f in args.rst_files]
 
+COMMON_ADC = """
```
gusthoff, line 222: For the moment, we're duplicating code from `backend/app/widget/project.py`.

Robert-Tice (approved):

PR #544, opened 2021-04-30
Editorial change: removing reference to obsolete documents.

gusthoff:


gusthoff:
Integrating this change into #545.

PR #557, opened 2021-05-22
Adding section on deferred constants.

gusthoff:


content/courses/advanced-ada/chapters/types.rst
```diff
@@ -945,13 +945,117 @@ default values, since we haven't assigned any value to them.
...
+       --      ^ full constant declaration using a subtype
+
+    end Deferred_Constants;
+
+Here, we're using the :ada:`Speed` type in the deferred declaration of the
+:ada:`Light` constant, but we're using the :ada:`Positive_Speed` subtype in
+the full declaration.
+
+A useful application of deferred constants is when the value of the constant is
+calculated in a function instead of being a literal. In this case, the full
```
pat-rogers, line 1034: Maybe emphasize the rationale:
"calculated using entities not meant to be compile-time visible to clients. As such, these other entities are only visible in the private part of the package, so that's where the value of the deferred constant must be computed. For example, the full constant declaration may be computed by a call to an expression function..."

gusthoff: Thanks for your suggestion. I've just integrated it.

pat-rogers (approved):

PR #564, opened 2021-06-05
Adding section on user-defined literals.

gusthoff:


content/courses/advanced-ada/chapters/types.rst
```diff
@@ -1178,13 +1178,350 @@ declaration of the :ada:`Light` constant.
...
+:ada:`Activation_State` type. It's our job to process that string parameter in
+the implementation of the :ada:`Integer_To_Activation_State` function and
+convert it to an integer value |mdash| using :ada:`Integer'Value`, for example:
+
+.. code:: ada manual_chop compile_button project=Courses.Advanced_Ada.Types.User_Defined_Literals
+
+    !activation_states.adb
+    package body Activation_States is
+
+       function Integer_To_Activation_State (S : String)
```
pat-rogers, line 1283: I think we should be consistent with the capitalization for the type name String. It is "string" in several places above, but also "String" too. (I'd advocate for "String".)

PR #579, opened 2021-07-03
Adding section on record representation clauses.

gusthoff:


content/courses/advanced-ada/chapters/types.rst
```diff
@@ -1212,16 +1212,434 @@ Data Representation
...
+       Put_Line ("R'Object_Size: " & R'Object_Size'Image);
+    end Show_Empty_Byte;
+
+If the code compiles, :ada:`R'Size` and :ada:`R'Object_Size` should now have
+the same value.
+
+Derived Types And Representation Clauses
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In some cases, you might want to modify the memory representation of a record
```
pat-rogers, line 1497: A common case for doing this is when you want to convert between the parent type and the derived type with a different representation. For example, one type might have a representation that is laid out for efficient access but therefore includes padding between some of the components.  We could have a compacted representation in the derived type, and use that to store the data on some external media. Using record representations and type conversions makes that easy.

gusthoff: Yes, I agree, this is an interesting application. Actually, this topic is discussed in Robert Dewar's Gems 27 and 28, which are very good and therefore were included in this PR: 

https://github.com/AdaCore/learn/pull/575/files#diff-b2fee03b65a4b060b1eeede8faf1e0ba657e0b4971889376e1c54fb8546a134aR2028

That section (containing Robert's gems) is supposed to come after this one, so the goal here is to just provide a short introduction into representation clauses and leave the more interesting details to be discussed in Robert's article.

pat-rogers (approved):

PR #584, opened 2021-07-09
Adding section on unchecked unions.

gusthoff:


content/courses/advanced-ada/chapters/types.rst
```diff
@@ -1239,15 +1239,240 @@ Valid
 Unchecked Union
 ---------------
 
-.. admonition:: Relevant topics
+We've introduced variant records back in the
+:doc:`Introduction to Ada course <courses/intro-to-ada/chapters/more_about_types>`.
+In simple terms, a variant record |mdash| or *discriminated record* in Ada
+terminology |mdash| is a record with discriminants that allows for changing its
```
pat-rogers, line 1245: But a discriminated record doesn't necessarily have a variant part, and often doesn't, so I would argue that they are not the same thing even if that is one of the primary motivations.

content/courses/advanced-ada/chapters/types.rst
```diff
@@ -1239,15 +1239,240 @@ Valid
...
+       V := (Use_Enum => False, I => 10);
+       Display_State_Value (V);
+    end Show_Unchecked_Union;
 
-    Complete section!
 
+However, in general, you should avoid using the :ada:`Unchecked_Union` aspect
+due to the potential issues you might introduce into your application. In most
+cases, unless you're doing low-level programming or interfacing with C code,
+you won't really need this aspect.
```
pat-rogers, line 1475: So much so that I don't think we should be telling them about it at all...

And I disagree that it is needed in low-level programming, the svd2ada generator notwithstanding. :-)

content/courses/advanced-ada/chapters/types.rst
gusthoff: Actually, this first phrase was taken from the "Ada for the Embedded C Dev." course. I was curious how this came to be and I found this comment:

https://github.com/AdaCore/learn/pull/404#discussion_r498873685

The reason I'm mentioning this here is because a correction is (most probably) needed in that course as well. Should I just remove the part about "record with discriminants" in both locations?

content/courses/advanced-ada/chapters/types.rst
gusthoff: I see what you mean, and I do agree that this topic is much less important than other topics I have on my list. 

My goal, however, is to provide a simple and short explanation about aspects that an Ada developer might encounter when reading someone else's code. Developers should be able to get a rudimentary understanding of what unchecked unions are by consulting this "Advanced Ada" course, even if they're not supposed / expected to use them. Unfortunately, reading the RM-B-3-3 section isn't an option in this case, as the text is too technical there.

content/courses/advanced-ada/chapters/types.rst
pat-rogers: No, I'm being too picky again. :-)

pat-rogers (approved):

PR #612, opened 2021-09-04
Adding section on scalar types

gusthoff:


content/courses/advanced-ada/chapters/types.rst
```diff
@@ -6,14 +6,432 @@ Types
...
+
+- the :ada:`Wide_Wide_Width` attribute for strings returned by
+  :ada:`Wide_Wide_Image`.
+
+Base
+~~~~
+
+The :ada:`Base` attribute gives us the unconstrained version of a subtype. As
+an example, let's say we declared a subtype of the :ada:`Integer` type named
+:ada:`One_To_Ten`:
```
pat-rogers, line 286: This is a subtle topic! 
Background: 
"Integer" is actually a subtype too. (These used to be known as "first named subtypes" -- the terminology has changed -- but the term does suggest correctly that they are a subtype of something that does not have an Ada name, and that is not constrained. That anonymous "type" is the hardware's representation for the numeric type as chosen by the compiler based on the requested range (for the signed integer types) or digits of precision (for floating-point types). Let's focus on signed integers. The other numerics work the same way. When we declare a signed integer type we have to specify the required range, statically. If the compiler cannot find a hardware-defined/supported signed integer type with at least the range requested, the compilation is rejected. Otherwise, the compiler maps the named Ada type to the hardware "type", presumably choosing the smallest one that supports the requested range. (That's why the range has to be static in the source code, unlike for explicit subtypes.) For any integer type named Foo, we can refer to that anonymous hardware type via Foo'Base.
(I've never used the term "first named subtype" in a training course and don't recommend it here either.) In effect, 

   type Integer is range L .. U;

is really as if we said this:

   subtype Integer is Some_Hardware_Type_With_Sufficient_Range range L .. U;

But that's impossible, so we just say that Integer is a type rather than a subtype.

So, what I suggest is 
"The :ada:`Base` attribute gives us the unconstrained version of a subtype." =>
"The :ada:`Base` attribute gives us the unconstrained underlying hardware representation selected for a given numeric type."

content/courses/advanced-ada/chapters/types.rst
```diff
@@ -6,14 +6,432 @@ Types
...
+    package My_Integers is
+
+       subtype One_To_Ten is Integer range 1 .. 10;
+
+    end My_Integers;
+
+If we then use the :ada:`Base` attribute |mdash| by writing
+:ada:`One_To_Ten'Base` |mdash|, we're actually referring to the
+:ada:`Integer` type (because the :ada:`Integer` type was used as the base
+subtype of the :ada:`One_To_Ten` subtype).
```
pat-rogers, line 299: Subtypes just introduce names for constraints on existing types or subtypes. So whenever we see "One_To_Ten" we are really working with type Integer. The only difference is the values returned for 'First and 'Last. You could say "One_To_Ten'Succ (One_To_Ten'Last)" and assign that value of 11 that to an Integer variable, for example. So, One_To_Ten'Base refers to the same base type as Integer'Base, which is the underlying hardware type representing type Integer but is not Integer itself.

content/courses/advanced-ada/chapters/types.rst
gusthoff: I've integrated this change you suggested. However, your explanation about the hardware representation is interesting, so I'm asking myself whether we should add it to the section (maybe as an admonition). What do you think?

content/courses/advanced-ada/chapters/types.rst
gusthoff: Thanks! I reworked this paragraph and tried to integrate your comments.

content/courses/advanced-ada/chapters/types.rst
pat-rogers: I agree that it would be good to have the material included.

gusthoff: Awesome! I added your explanation to the section and did some modifications here and there.

pat-rogers (approved):

pat-rogers:
Looks good!

PR #637, opened 2021-10-29
Introducing section on floating-point and fixed-point attributes

gusthoff:


gusthoff:
I've just reworked this pull request. Let me know if you have further remarks and suggestions.

gusthoff:
Here are some comments from my side:

> *_Rounding, you say it returns the nearest integer value, but the whole
> point is that if you are halfway between, it will return one of the nearest
> integer values and it is just a question of which one. [I realize the RM
> makes this same confusing use of "the" but no need for us to repeat that
> confusion.]

I'm actually mentioning this in the section:

    The difference between these attributes is the way they handle the case when a
    value is exactly in between two integer values. For example, 4.5 could be
    rounded up to 5.0 or rounded down to 4.0.

However, maybe I'm misunderstanding your comment.


> Also, why do you give the result of Adjacent(1.0e-83, 0.0) is -0.0? It
> should be +0.0.

I adapted the text. However, that's the value I actually get when running the code snippet.

gusthoff:
@sttaft: I've just reworked this pull request. Let me know if you have further remarks and suggestions.

content/courses/advanced-ada/chapters/numerics.rst
```diff
@@ -963,12 +998,12 @@ Attributes: :ada:`'Copy_Sign` and :ada:`Leading_Part`
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 :ada:`'Copy_Sign` is an attribute that returns a value where the sign of the
-second floating-point argument is applied to the first floating-point argument.
-For example, :ada:`Float'Copy_Sign (1.0, -10.0)` is -1.0. Here, the sign of the
-second argument (-10.0) is applied to the first argument (1.0), so the result
-is -1.0.
+second floating-point argument is multiplied to the magnitude of the first
```
sttaft, line 1001: ... multiplied *by* the magnitude ...

content/courses/advanced-ada/chapters/numerics.rst
```diff
@@ -963,12 +998,12 @@ Attributes: :ada:`'Copy_Sign` and :ada:`Leading_Part`
...
+floating-point argument. For example, :ada:`Float'Copy_Sign (1.0, -10.0)` is
+-1.0. Here, the sign of the second argument (-10.0) is multiplied to the
```
sttaft, line 1003: ... is multiplied *by* the ...

sttaft (approved):

PR #642, opened 2021-11-12
Adding sections on expression functions, conditional expressions and quantified expressions

gusthoff:


content/courses/advanced-ada/chapters/subprograms.rst
```diff
@@ -6,21 +6,456 @@ Subprograms
...
+block and we have at least one :ada:`return` statement in that block. An
+expression function, in contrast, is a function that is implemented with a
+simple expression in parentheses, such as :ada:`(X);`. In this case, we don't
+use a :ada:`begin ... end;` block or a :ada:`return` statement.
 
-    Complete section!
+As an example of an expression, let's say we want to implement a function
+named :ada:`Is_Zero` that checks if the value of the integer parameter :ada:`I`
+is zero. We can implement this function with the expression :ada:`I = 0`. In
+the standard approach, we would create the implementation by writing
```
pat-rogers, line 19: "standard" suggests more than it should. Maybe use "usual" again, to tie it to the previous paragraph?

content/courses/advanced-ada/chapters/subprograms.rst
```diff
@@ -6,21 +6,456 @@ Subprograms
...
+.. code:: ada compile_button project=Courses.Advanced_Ada.Numerics.Simple_Expression_Function_1
+
+    package Expr_Func is
+
+       function Is_Zero (I : Integer) return Boolean is
+         (I = 0);
+
+    end Expr_Func;
+
+An expression function has the same effect as the standard, more verbose
```
pat-rogers, line 34: I suggest cutting "standard, more verbose" altogether, or take out "standard" anyway. The phrase "more verbose" is accurate, but too pejorative IMO.

content/courses/advanced-ada/chapters/subprograms.rst
```diff
@@ -6,21 +6,456 @@ Subprograms
...
+
+Conditional Expressions
+-----------------------
+
+In the previous section, we've seen simple expressions such as :ada:`I = 0` or
+:ada:`D.Valid`. A conditional expression, as the name implies, is an
+expression that contains a condition. This might be an "if-expression" (in the
+:ada:`if ... then ... else` form) or a "case-expression" (in the
+:ada:`case ... is when =>` form). The result of a conditional expression is
+always a Boolean value.
```
pat-rogers, line 161: "The result of a conditional expression is always a Boolean value."  That's not right, if I understand the intent. The result is a value of a type determined by the context. For example, the result of the conditional expression implementing function Max is Integer, because that's the type that Max is declared to return. The values of A and B are checked to ensure they are Integer values.

content/courses/advanced-ada/chapters/subprograms.rst
gusthoff: I'll use "usual" instead as you suggested above.

content/courses/advanced-ada/chapters/subprograms.rst
gusthoff: You're absolutely right, of course! Even my example contradicts this statement, so I'll just remove it.

(I'm not sure how this ended up here, but I guess this is just a copy-and-paste error, as this phrase is part of the next subsection on quantified expressions.)

pat-rogers (approved):

PR #656, opened 2021-12-17
Adding section on nonreturning procedures

gusthoff:


content/courses/advanced-ada/chapters/subprograms.rst
```diff
@@ -724,13 +724,144 @@ Operator Overloading
 Nonreturning procedures
 -----------------------
 
-.. admonition:: Relevant topics
+Usually, when calling a procedure, we expect that it returns after a
+reasonable amount of processing time. However, there are situations where a
```
pat-rogers, line 728: I suggest you consider emphasizing the return to the caller's "thread of control," and that it does so after performing some action, rather than after some amount of time.

gusthoff: Thanks for the suggestion, I reused it in the paragraph.

pat-rogers (approved):

PR #657, opened 2021-12-17
Adding section on inlining subprograms

gusthoff:


content/courses/advanced-ada/chapters/subprograms.rst
```diff
@@ -736,14 +736,110 @@ Nonreturning procedures
 Inline subprograms
 ------------------
 
-.. admonition:: Relevant topics
+`Inlining <https://en.wikipedia.org/wiki/Inline_expansion>`_ refers to a kind
+of optimization where the code of a subprogram is expanded instead of being
```
pat-rogers, line 740: How about instead say "is expanded at the point of the call instead of the call itself." or something like that which emphasizes the replacement of the call.

pat-rogers (approved):

PR #664, opened 2022-01-07
Adding section on simple and compound statements

gusthoff:


content/courses/advanced-ada/chapters/statements.rst
```diff
@@ -3,6 +3,133 @@ Statements
...
+
+.. admonition:: In the Ada Reference Manual
+
+    - `5.1 Simple and Compound Statements - Sequences of Statements <http://www.ada-auth.org/standards/12rm/html/RM-5-1.html>`_
+
+Labels
+~~~~~~
+
+We can use labels to identify statements in the code. They are useful as a way
+to document the code or to indicate important statements in the code.
```
pat-rogers, line 32: I disagree with this claim. I've never seen or heard of them being used for those two purposes. They are only in the language, as far as I know, for being the targets of goto statements. If I came across one that wasn't a goto target I'd think it was very odd indeed.
One can make the case that goto statements have benefit as a way to construct the effect of "continue" statements in other languages, which Ada is missing.

content/courses/advanced-ada/chapters/statements.rst
```diff
@@ -3,6 +3,133 @@ Statements
...
+
+    with Ada.Text_IO; use Ada.Text_IO;
+
+    procedure Show_Statement_Identifier is
+       pragma Warnings (Off, "is not referenced");
+
+       Arr   : constant array (1 .. 5) of Integer := (1, 4, 6, 42, 49);
+       Found : Boolean := False;
+    begin
+       <<Find_42>> for E of Arr loop
```
pat-rogers, line 72: Seems better as a loop identifier, IMO.

content/courses/advanced-ada/chapters/statements.rst
```diff
@@ -3,6 +3,133 @@ Statements
...
+              if E = 42 then
+                 Found := True;
+                 exit Find_42;
+              end if;
+           end loop Find_42;
+
+           Put_Line ("Found: " & Found'Image);
+        end Show_Statement_Identifier;
+
+    Later in this chapter, we discuss this topic in more details.
```
pat-rogers, line 109: "details" should be "detail"

content/courses/advanced-ada/chapters/statements.rst
```diff
@@ -3,6 +3,133 @@ Statements
...
+       Some_Error := False;
+
+       if Some_Error then
+          goto Cleanup;
+       end if;
+
+       <<Cleanup>> null;
+    end Show_Cleanup;
+
+Here, we transfer the control to the *cleanup* statement as soon as an error is
```
pat-rogers, line 129: And here I'd suggest a description of using a goto statement in a loop to implement a continue statement, which can simplify the number of boolean flags required for loop termination in some cases.

content/courses/advanced-ada/chapters/statements.rst
gusthoff: Thanks for the insight. I changed the text to focus more on `goto` statements.

content/courses/advanced-ada/chapters/statements.rst
gusthoff: Agreed. My goal is to just show that using labels is also possible in this case. However, I added this phrase:

> Loop statement and block statement identifiers are generally preferred over labels.

content/courses/advanced-ada/chapters/statements.rst
gusthoff: Usually, I hardly need `goto` statements, so I'm not sure I've ever encountered this use case where a `goto` would be preferable in a loop. Could you please show me what code you have in mind?

pat-rogers: I've only used it once, in an executive:   <<spin>> goto spin;

They are rife throughout the gnat compiler code. The simplest I could find is this one:

   procedure Qualify_All_Entity_Names is
      E   : Entity_Id;
      Ent : Entity_Id;
      Nod : Node_Id;

   begin
      for J in Name_Qualify_Units.First .. Name_Qualify_Units.Last loop
         Nod := Name_Qualify_Units.Table (J);

         --  When a scoping construct is ignored Ghost, it is rewritten as
         --  a null statement. Skip such constructs as they no longer carry
         --  names.

         if Nkind (Nod) = N_Null_Statement then
            goto Continue;
         end if;

         E := Defining_Entity (Nod);
         Reset_Buffers;
         Qualify_Entity_Name (E);

         --  Normally entities in the qualification list are scopes, but in the
         --  case of a library-level package renaming there is an associated
         --  variable that encodes the debugger name and that variable is
         --  entered in the list since it occurs in the Aux_Decls list of the
         --  compilation and doesn't have a normal scope.

         if Ekind (E) /= E_Variable then
            Ent := First_Entity (E);
            while Present (Ent) loop
               Reset_Buffers;
               Qualify_Entity_Name (Ent);
               Next_Entity (Ent);

               --  There are odd cases where Last_Entity (E) = E. This happens
               --  in the case of renaming of packages. This test avoids
               --  getting stuck in such cases.

               exit when Ent = E;
            end loop;
         end if;

         <<Continue>>
         null;
      end loop;
   end Qualify_All_Entity_Names;

However the better case is in while-loops, where "implementing" a continue statement via the goto allows one not need another boolean variable to control the loop iteration. 

All that said, I think refactoring to simplify the loop content is a better approach. I only mention the "continue statement" stuff in case an example is needed.

pat-rogers (approved):

content/courses/advanced-ada/chapters/statements.rst
gusthoff: Thanks, now I understand what you mean.

I just wrote a small subsection that explains the basic idea behind a `Continue` label.

pat-rogers (approved):

PR #678, opened 2022-02-06
Adding section on Out and Uninitialized

gusthoff:


content/courses/advanced-ada/chapters/exceptions.rst
```diff
@@ -247,9 +247,172 @@ Out and Uninitialized
...
+        end Show_Out_Uninitialized_Warnings;
+
+    We now get a compilation warning that the pass-by-copy formal may have no
+    effect.
+
+    Of course, GNAT is not able to point out all such errors (see first example
+    above), which in general would require full flow analysis.
+
+The behavior is different when using parameter types that the standard mandates
+passing by reference, such as tagged types for instance. So the following code
```
pat-rogers, line 323: "The behavior is different when using parameter types that the standard mandates passing by reference, such as tagged types for instance." is a little awkward (but not much). 
Maybe:
"The behavior is different when using parameter types that the standard mandates be passed by reference, such as tagged types for instance."
Or:
"The behavior is different when using parameter types that are passed by reference, such as tagged types for instance."

pat-rogers (approved):

PR #707, opened 2022-04-03
Fix RST markup

reznikmm:


gusthoff (approved):
Good catch! Thanks for fixing this issue.

PR #712, opened 2022-04-08
Adding section on limited visibility

gusthoff:


content/courses/advanced-ada/chapters/packages.rst
```diff
@@ -181,16 +181,197 @@ possible to access the information from a private package in the private part
...
+          Value : T1;
+       end record;
+
+    end B;
+
+Here, we have two
+:ref:`mutually dependent types <Adv_Ada_Mutually_Dependent_Types>` :ada:`T1`
+and :ada:`T2`, which are declared in two packages :ada:`A` and :ada:`B` that
+refer to each other. These with clauses constitute a circular dependency, so
+the compiler cannot compile any of those packages.
```
pat-rogers, line 218: Consider changing "any" to "either" because "any" suggests several, whereas there are precisely two in this scenario...

content/courses/advanced-ada/chapters/packages.rst
```diff
@@ -181,16 +181,197 @@ possible to access the information from a private package in the private part
...
+
+    package A is
+
+       type T1 is record
+          Ref : access B.T2;
+       end record;
+
+    end A;
+
+    limited with A;
```
pat-rogers, line 264: Note that it is OK to have a limited-with on both specs, but not necessary; you only need the limited-with on one of the two specs. So in package B, for example, you can refer directly to B.T2 as the record component type without the anonymous access type.

content/courses/advanced-ada/chapters/packages.rst
```diff
@@ -181,16 +181,197 @@ possible to access the information from a private package in the private part
...
+
+    end B;
+
+As expected, the code can now be compiled without issues.
+
+.. admonition:: In the Ada Reference Manual
+
+    - `10.1.2 Context Clauses - With Clauses <http://www.ada-auth.org/standards/12rm/html/RM-10-1-2.html>`_
+
+Limited view and private with clauses
```
pat-rogers, line 280: "Limited view" is a little problematic because of limited types, which provide a limited view...

content/courses/advanced-ada/chapters/packages.rst
```diff
@@ -181,16 +181,197 @@ possible to access the information from a private package in the private part
...
+       --  Here, we have limited visibility
+       --  of package B
+
+       type T1 is record
+          Ref : access B.T2;
+       end record;
+
+    end A;
+
+    limited private with A;
```
pat-rogers, line 309: Same comment for this as above: only one is required.

content/courses/advanced-ada/chapters/packages.rst
gusthoff: Thanks, I've just corrected it.

content/courses/advanced-ada/chapters/packages.rst
gusthoff: Good point, thanks for the hint! I've just changed both code examples, and added a new one that shows this use-case of having `limited with` on both packages.

content/courses/advanced-ada/chapters/packages.rst
gusthoff: I found the term "limited view" in the RM 10.1.2 [17/2] itself. In any case, I've replaced the term everywhere with "limited visibility", which is hopefully less confusing.

pat-rogers (approved):

PR #730, opened 2022-05-27
Adding sections on use type clause and on naming conflicts

gusthoff:
Adding sections:
- "Use type clause"
- "Use clauses and naming conflicts"


gusthoff:
Thanks for all the comments! I adapted the text accordingly. Also, I changed similar usages in other (older) sections.

PR #745, opened 2022-06-25
Adding comments on well-defined behavior of Unchecked_Conversion

### LINKED ISSUE #504: Review Unchecked_Conversion for correctness in Ada for the C developer
In this chapter of the book: https://learn.adacore.com/courses/Ada_For_The_Embedded_C_Developer/chapters/06_Translation.html#

the following statement is made:
"Also note that Unchecked_Conversion can only be instantiated for constrained types"

However, per the standard http://www.ada-auth.org/standards/12aarm/html/AA-13-9.html
"The size of the formal parameter S in an instance of Unchecked_Conversion is that of its subtype. [This is the actual subtype passed to Source, except when the actual is an unconstrained composite subtype, in which case the subtype is constrained by the bounds or discriminants of the value of the actual expression passed to S.]"

I have written a reproducer with Unchecked_Conversion instantiated for an unconstrained array which does compile:
```
begin
   declare
      type Payload is record
         I : Integer;
         F : Float;
      end record;

      function Process (Data : String) return Payload is
         subtype Input_String is String (Data'First .. Data'Last);
         function Convert is new Ada.Unchecked_Conversion (Source => Input_String, Target => Payload);
--           Heap_Data  : String := new String'(Data);
--           Return_Value : Payload with Address => Data'Address;
         Return_Value : Payload;
      begin
--           Put_Line ( Data'Size'Image );
--           Put_Line ( Convert (Data)'Size'Image);
--           return Convert (Data);
         Return_Value := Convert ( Data );
         Put_Line (Return_Value.F'Valid'Image);
         return Return_Value;
      end Process;
      P : Payload;
   begin
      P := Process ("CATS ARE AWESOME");
      Put_Line (P.I'Image);
      Put_Line (P.F'Image);
   end;
end Main;
```

gusthoff:
Fix #504

### DIFF
--- a/content/courses/Ada_For_The_Embedded_C_Developer/chapters/06_Translation.rst
+++ b/content/courses/Ada_For_The_Embedded_C_Developer/chapters/06_Translation.rst
@@ -1548,8 +1548,29 @@ in the source object is automatically reflected in the target object (and
 vice-versa). In the end, the choice between unchecked conversions and overlays
 depends on the level of performance that you want to achieve.
 
-Also note that :ada:`Unchecked_Conversion` can only be instantiated for
-constrained types. In order to rewrite the examples using bit-fields that we've
+Also note that :ada:`Unchecked_Conversion` will only have defined behavior
+when instantiated for constrained types. For example, we shouldn't use this
+kind of conversion:
+
+.. code-block:: ada
+
+   Ada.Unchecked_Conversion (Source => String,
+                             Target => Integer);
+
+Although this compiles, the behavior will only be well-defined in those cases
+when :ada:`Source'Size = Target'Size`. Therefore, instead of using an
+unconstrained type for :ada:`Source`, we should use a subtype that matches this
+expectation:
+
+.. code-block:: ada
+
+   subtype Integer_String is String (1 .. Integer'Size / Character'Size);
+
+   function As_Integer is new
+     Ada.Unchecked_Conversion (Source => Integer_String,
+                               Target => Integer);
+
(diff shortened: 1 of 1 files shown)


content/courses/Ada_For_The_Embedded_C_Developer/chapters/06_Translation.rst
```diff
@@ -1548,8 +1548,29 @@ in the source object is automatically reflected in the target object (and
 vice-versa). In the end, the choice between unchecked conversions and overlays
 depends on the level of performance that you want to achieve.
 
-Also note that :ada:`Unchecked_Conversion` can only be instantiated for
-constrained types. In order to rewrite the examples using bit-fields that we've
+Also note that :ada:`Unchecked_Conversion` will only have defined behavior
```
pat-rogers, line 1551: A side comment, not important in itself. Since the name of the generic is Ada.Unchecked_Conversion, and since Unchecked_Conversion is still a valid unit name, I suggest you say "unchecked conversion" when you're talking about the semantic effect rather than the unit name.

pat-rogers (approved):

content/courses/Ada_For_The_Embedded_C_Developer/chapters/06_Translation.rst
gusthoff: Yes, I agree. I've just changed a few instances of it in the text.

pat-rogers (approved):

PR #747, opened 2022-06-25
Adding simple explanation about .ads/.adb files

### LINKED ISSUE #578: .ads files not explained at the point they are introduced
I hope you don't mind if I post my impressions as I go. As someone who has written docs, I'm probably more nit-picky than most. If my comments aren't useful, please let me know.

At the point you introduce .ads files, there's no explanation of what they are. For anyone who hasn't used Modula, their role and usage is not obvious. I had to google to put them into some context. 

https://learn.adacore.com/courses/intro-to-ada/chapters/subprograms.html

To the new user from languages like Python, C# & Java they might look like an irritation. A crisp explanation of their advantages to both the compiler and the developer might help cushion the blow.

gusthoff:
Fix #578

### DIFF
--- a/content/courses/intro-to-ada/chapters/subprograms.rst
+++ b/content/courses/intro-to-ada/chapters/subprograms.rst
@@ -73,7 +73,7 @@ This is the implementation of the function above:
    the :ada:`Increment` function above in a file called :file:`increment.ada`.
    From the standard's perspective, this would be completely acceptable.
 
-   The GNAT toolchain, however, requires the following file structure:
+   The GNAT toolchain, however, requires the following file naming scheme:
 
    - files with the `.ads` extension contain the specification, while
 
--- a/content/courses/intro-to-ada/chapters/subprograms.rst
+++ b/content/courses/intro-to-ada/chapters/subprograms.rst
@@ -64,6 +64,32 @@ This is the implementation of the function above:
        return I + Incr;
     end Increment_By;
 
+.. admonition:: In the GNAT toolchain
+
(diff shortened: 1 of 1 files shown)


content/courses/intro-to-ada/chapters/subprograms.rst
```diff
@@ -64,6 +64,32 @@ This is the implementation of the function above:
        return I + Incr;
     end Increment_By;
 
+.. admonition:: In the GNAT toolchain
+
+   The Ada standard doesn't mandate in which file the specification or the
+   implementation of a subprogram must be stored. In other words, the standard
+   doesn't require a specific file structure or specific file name extensions.
+   For example, we could save both the specification and the implementation of
+   the :ada:`Increment` function above in a file called :file:`increment.ada`.
```
pat-rogers, line 73: Maybe go further, to make the point, and use "increment.txt" rather than an extension that at least looks like a source file name?
You can have the entire source code in one file if you want. Dumb, but legal.
For that matter, from the language's point of view, you could use some other representation instead of a file.

content/courses/intro-to-ada/chapters/subprograms.rst
```diff
@@ -64,6 +64,32 @@ This is the implementation of the function above:
...
+   From the standard's perspective, this would be completely acceptable.
+
+   The GNAT toolchain, however, requires the following file structure:
```
pat-rogers, line 76: Instead of "structure" consider "naming scheme"

PR #752, opened 2022-07-01
Adding section on Unchecked Deallocation

gusthoff:


content/courses/advanced-ada/chapters/access_types.rst
```diff
@@ -294,14 +294,289 @@ Unchecked Access
...
+       procedure Free is
+         new Ada.Unchecked_Deallocation
+           (Object => Integer,
+            Name   => Integer_Access);
+    end Integer_Types;
+
+Here, we declare the :ada:`Free` procedure, which we can then use to deallocate
+objects that were allocated for the :ada:`Integer_Access` type.
+
+:ada:`Unchecked_Deallocation` is a generic procedure that we can instantiate
```
pat-rogers, line 326: I suggest :ada:`Ada.Unchecked_Deallocation` throughout, to avoid confusion with the old name that is still legal and available.

content/courses/advanced-ada/chapters/access_types.rst
```diff
@@ -294,14 +294,289 @@ Unchecked Access
...
+       I := new Integer;
+
+       Free (I);
+       Free (I);
+       Free (I);
+    end Show_Unchecked_Deallocation;
+
+The multiple calls to :ada:`Free` in this example don't cause any issues. In
+fact, calling :ada:`Free` to an access value that is equal to :ada:`null` is
+simply ignored.
```
pat-rogers, line 460: Do you want to talk about the danger of calling Free on the same designated object using another access value also designating that same object?

content/courses/advanced-ada/chapters/access_types.rst
```diff
@@ -294,14 +294,289 @@ Unchecked Access
...
+the corresponding access value is set to :ada:`null`. In this case, we can
+check for :ada:`null` to detect a dangling reference and avoid dereferencing
+it.
+
+The strategy of checking for :ada:`null` before trying to dereferencing the
+access value works fine for :ada:`I_1`, but not for :ada:`I_2`, as we can see
+when running the example above. In the case of :ada:`I_2`, the access value is
+not :ada:`null`: :ada:`I_2` points to an object that doesn't exist anymore. As
+expected, dereferencing :ada:`I_2` raises the :ada:`Constraint_Error`
+exception.
```
pat-rogers, line 520: The evaluation of a dangling reference is erroneous, meaning the behavior in that case is undefined. If the compiler can detect the situation it can raise C_E, but more likely it will just give junk results for the dereference, or worse yet appear to work correctly if the junk is a valid value, or launch rockets, or anything else. This issue is discussed in RM 13.9.1 (13) and especially 13.11.2 (15).

content/courses/advanced-ada/chapters/access_types.rst
gusthoff: I added a new subsection on dangling references (after this one) that discuss calls to `Free` when two access values are used.

gusthoff: Just let me know if I should add another example.

content/courses/advanced-ada/chapters/access_types.rst
gusthoff: Good point! The code example was actually wrong. (This is now corrected.) Also, I rewrote this part to emphasize that the behavior is undefined.

content/courses/advanced-ada/chapters/access_types.rst
```diff
@@ -294,14 +294,402 @@ Unchecked Access
...
+       if I_2 /= null then
+          Put_Line ("I_2 = " & Integer'Image (I_2.all));
+       end if;
+    end Show_Unchecked_Deallocation;
+
+In this example, we allocate an object for :ada:`I_1` and make :ada:`I_2` point
+to the same object. Note, however, that we call :ada:`Free` for :ada:`I_1`, but
+not for :ada:`I_2`. Therefore, after that call to :ada:`Free`,
+:ada:`I_1 = null`, but :ada:`I_2` still points to the object originally
+allocated for :ada:`I_1`.
```
pat-rogers, line 590: No, I_2 may or may not point to the originally allocated object after the call to Free to I_1. I_2 still has the non-null access value that formerly designated I_1.all, but I_1.all has been reclaimed by the call to Free and may be reused for other allocations. So I_2.all is non-null but what it points to is now undefined.
All we can say is that after the call to Free for I_1, I_2 remains non-null but becomes a dangling reference, and the effect of dereferencing I_2 is undefined. And really difficult to debug because the storage at the old address of I_1.all might not yet be used by another allocation, so it could look like I_1.all (and hence I2.all) is "still there" with the expected values.

content/courses/advanced-ada/chapters/access_types.rst
```diff
@@ -294,14 +294,402 @@ Unchecked Access
...
+
+As mentioned previously, when deallocating an object via a call to :ada:`Free`,
+the corresponding access value is set to :ada:`null`. In this case, we can
+check for :ada:`null` to detect a dangling reference and avoid dereferencing
+it.
+
+The strategy of checking for :ada:`null` before trying to dereferencing the
+access value works fine for :ada:`I_1`. However, this strategy doesn't work for
+:ada:`I_2`: the access value is not :ada:`null` because it still points to an
+object that existed at some point. Therefore, the application will try to
```
pat-rogers, line 600: Yes, exactly right, so what I am suggesting is careful wording in the earlier text.

content/courses/advanced-ada/chapters/access_types.rst
```diff
@@ -294,14 +294,402 @@ Unchecked Access
...
+points to):
+
+.. code-block:: ada
+
+       Put_Line ("Freeing I_1");
+       I_2 := null;
+       Free (I_1);
+
+With this fix, :ada:`I_2` isn't a dangling reference anymore after the call to
+:ada:`Free (I_1)`.
```
pat-rogers, line 637: It won't be practical for the programmer to make every possible dangling reference become null if the calls to Free are strewn throughout the code. That means the suggested design is to not use Free in client code, but instead hide its use within the bigger abstraction.  That way all the occurrences of the calls to Free are in one package, and the programmer of that package then can prevent dangling references. For example, if we have a bank account that involves dynamic allocation for some part of an account, the bank account is a private type (or limited), and the access values are hidden in the private part. Thus when an account is closed the allocated memory can be freed without the possibility of leaving dangling references. That will work as long as the reference cannot be copied when copying bank account objects, in which case we'd want the type to be limited as well as private. Or we'd make it a controlled type so that Finalize can call Free. I don't know if you will want to mention this design idiom, and I'm not sure  a bank account is a good example! :-)

gusthoff:
Hey @pat-rogers, thanks a lot for all the comments. This was extremely helpful! 

I reworked some of the sections. Let's see if I got it right this time. :-)

(I used some of your comments in this new version; I hope this is OK with you.)

content/courses/advanced-ada/chapters/access_types.rst
gusthoff: Thanks a lot for this comment! I agree, we should describe good design patterns for access types in a separate section. (I added a todo item for that, it's on my list now.)

content/courses/advanced-ada/chapters/access_types.rst
```diff
@@ -294,14 +294,430 @@ Unchecked Access
...
+       --  A call to Free (I_2) is
+       --  NOT OK:
+
+       Free (I_2);
+    end Show_Unchecked_Deallocation;
+
+As we've seen before, we can have multiple calls to :ada:`Free (I)`.
+However, the call to :ada:`Free (I_2)` is bad because :ada:`I_2` is not null.
+In fact, it is a dangling reference |mdash| i.e. :ada:`I_2` points to an object
+that doesn't exist anymore. Also, the previous call to :ada:`Free (I)` will try
```
pat-rogers, line 524: 'Also, the previous call to " should probably be "Also, the first call to " because the previous call is the third call for I, which has no effect.

gusthoff: Good catch, thanks!

content/courses/advanced-ada/chapters/access_types.rst
```diff
@@ -521,7 +521,7 @@ same object as :ada:`I`:
 As we've seen before, we can have multiple calls to :ada:`Free (I)`.
 However, the call to :ada:`Free (I_2)` is bad because :ada:`I_2` is not null.
 In fact, it is a dangling reference |mdash| i.e. :ada:`I_2` points to an object
-that doesn't exist anymore. Also, the previous call to :ada:`Free (I)` will try
+that doesn't exist anymore. Also, the first call to :ada:`Free (I)` will try
```
pat-rogers, line 524: OK, good regarding "first".
One more thing. I should have mentioned this previously. Instead of "will try to reclaim" I think it should be "will reclaim". Then, when talking about Free (I_2) you can say "will try to reclaim the previously-reclaimed object" as the explanation for why the call is bad.

gusthoff: Yes, that makes totally sense. I hope I got it right now. :-)

pat-rogers (approved):

PR #763, opened 2022-08-05
Fix some code blocks

rowan-walshe:
Some code blocks were pulling in files that they shouldn't have been because they shared a project name with a code block that they shouldn't have.

Some tests were also failing due to missing failure expectations

### DIFF
--- a/content/courses/Ada_For_The_Embedded_C_Developer/chapters/02_Perspective.rst
+++ b/content/courses/Ada_For_The_Embedded_C_Developer/chapters/02_Perspective.rst
@@ -1768,7 +1768,7 @@ consider this C example:
 
 [C]
 
-.. code:: c manual_chop run_button project=Courses.Ada_For_Embedded_C_Dev.Perspective.Overflow_Wraparound_0
+.. code:: c manual_chop run_button project=Courses.Ada_For_Embedded_C_Dev.Perspective.Overflow_Wraparound_C
 
     !main.c
     #include <stdio.h>
@@ -1788,7 +1788,7 @@ The corresponding code in Ada raises an exception:
 
 [Ada]
 
-.. code:: ada run_button project=Courses.Ada_For_Embedded_C_Dev.Perspective.Overflow_Wraparound_1
+.. code:: ada run_button project=Courses.Ada_For_Embedded_C_Dev.Perspective.Overflow_Wraparound_Ada
     :class: ada-run-expect-failure
 
     with Ada.Text_IO; use Ada.Text_IO;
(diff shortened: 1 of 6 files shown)


content/courses/Ada_For_The_Embedded_C_Developer/chapters/02_Perspective.rst
```diff
@@ -1809,7 +1809,7 @@ a 32-bit modular type:
 
 [Ada]
 
-.. code:: ada run_button project=Courses.Ada_For_Embedded_C_Dev.Perspective.Overflow_Wraparound
+.. code:: ada run_button project=Courses.Ada_For_Embedded_C_Dev.Perspective.Overflow_Wraparound_2
```
gusthoff, line 1812: I agree with the distinction between the Ada and the C code. However, we should be able to use `Overflow_Wraparound_1` here as well (instead of giving it a new project name), as the current block will simply overwrite the previous one. Also, the `ada-run-expect-failure` class from the previous block shouldn't play a role here, as the class is block specific.

content/courses/Ada_For_The_Embedded_C_Developer/chapters/02_Perspective.rst
```diff
@@ -1768,7 +1768,7 @@ consider this C example:
 
 [C]
 
-.. code:: c manual_chop run_button project=Courses.Ada_For_Embedded_C_Dev.Perspective.Overflow_Wraparound
+.. code:: c manual_chop run_button project=Courses.Ada_For_Embedded_C_Dev.Perspective.Overflow_Wraparound_0
```
gusthoff, line 1771: Minor remark: the change is OK, but I'd prefer calling them `Overflow_Wraparound_C` and `Overflow_Wraparound_Ada`, as we're doing elsewhere — just to keep things consistent.

rowan-walshe: Happy to change them :+1:

content/courses/Ada_For_The_Embedded_C_Developer/chapters/02_Perspective.rst
rowan-walshe: When I change them I'll use the _Ada for both rather than creating two :+1: 

The reason for the need to have different C and Ada project names here change is because as it generates a `main.c` and `main.adb`, the two ada projects pick up this `main.c` and then gprbuild gives this error: `main.gpr:2:09: "main.adb" and "main.c" have the same object file name`

gusthoff (approved):

PR #775, opened 2022-08-22
Minor fixes in intro-to-ada

reznikmm:


content/courses/intro-to-ada/chapters/appendices.rst
```diff
@@ -155,52 +155,52 @@ Indefinite version
...
+| Indefinite limited private type                              | Any type, limited or     |
+|                                                              | not, indefinite or       |
+| **Format**: :ada:`type T (<>) is limited private;`           | definite                 |
++--------------------------------------------------------------+--------------------------+
+| Incomplete indefinite tagged private type                    | Any concrete tagged      |
+|                                                              | type,                    |
+| **Format**: :ada:`type T (<>) is tagged;`                    | indefinite or definite   |
++--------------------------------------------------------------+--------------------------+
+| Indefinite tagged private type                               | Any concrete, nonlimited |
+|                                                              | tagged type,             |
```
gusthoff, line 178: Editorial remark: instead of changing the whole table just because of a single change, I'd really prefer you just add a new line (after "Any concrete" and before "tagged type") to write "nonlimited".

content/courses/intro-to-ada/chapters/modular_programming.rst
```diff
@@ -199,7 +199,7 @@ user of the :ada:`Operations` package, providing a first form of encapsulation.
 This works because entities declared in the body are *only* visible in the
 body.
 
-This example shows how :ada:`Last_Increment` is used indirectly:
+This example shows how :ada:`Increment_By` is used indirectly:
```
gusthoff, line 202: `Last_Increment` was actually correct, as we're referring to the variable, not the function.

content/courses/intro-to-ada/chapters/privacy.rst
```diff
@@ -247,11 +247,11 @@ summarizes this:
...
 |                               | its parent's specification    |
 +===============================+===============================+
-| Specification: public part    |                               |
+| Specification: public part    | No                            |
 +-------------------------------+-------------------------------+
-| Specification: private part   | X                             |
+| Specification: private part   | Yes                           |
 +-------------------------------+-------------------------------+
-| Body                          | X                             |
+| Body                          | Yes                           |
```
gusthoff, line 254: This isn't really a "minor fix", but rather an editorial change. Having explicit "no"s and "yes"s might be unequivocal, but it also makes the tables more difficult to read.

I assume you have a problem with the "X", which could potentially be interpreted as "doesn't apply" in some contexts. If that's the case, you could replace them by check marks, which always mean "yes" — for that, use the `|check|` syntax from `isopub.txt`.

content/courses/intro-to-ada/chapters/appendices.rst
reznikmm: :+1:   it didn't cross my mind

content/courses/intro-to-ada/chapters/privacy.rst
reznikmm: > be interpreted as "doesn't apply"

This was exactly my case. `|check|` should work indeed.

gusthoff (approved):
Perfect, thanks a lot!

PR #776, opened 2022-08-25
Edit `Tasking` section of "Introducion to Ada"

reznikmm:


content/courses/intro-to-ada/chapters/tasking.rst
```diff
@@ -47,14 +47,17 @@ that both calls to :ada:`Put_Line` are performed.
...
-- The main application is itself a task (the main task).
+- The main application is itself a task (the main or “environment” task).
 
   - In this example, the subprogram :ada:`Show_Simple_Task` is the main task of
     the application.
 
 - Task :ada:`T` is a subtask.
 
-  - Each subtask has a master task.
+  - Each subtask has a master, which represents the program construct in which
```
gusthoff, line 57: Two comments here: 

1. There are four "which"s in this paragraph. Maybe you could rewrite this?
2. The term "construct" (as in "program construct" or "master construct") won't be understood by everyone. As it appears here for the first time, you probably should explain it in a sentence.

content/courses/intro-to-ada/chapters/tasking.rst
```diff
@@ -195,27 +200,28 @@ In this example, we're making the task :ada:`T` wait one second after each
...
 synchronize with it. Specifically,
 
-- The subtask waits at that point (in the :ada:`accept` statement),
+- The task with the entry waits at that point (in the :ada:`accept` statement),
   ready to accept a call to the corresponding entry from the master task.
 
-- The master task calls the task entry, in a manner similar to a procedure
-  call, to synchronize with the subtask.
+- The other task calls the task entry, in a manner similar to a procedure
+  call, to synchronize with it.
```
gusthoff, line 222: "it" can be confusing in this context. We should be more specific (as in the original version).

content/courses/intro-to-ada/chapters/tasking.rst
```diff
@@ -246,10 +252,11 @@ example:
...
 In this example, we declare an entry :ada:`Start` for task :ada:`T`.  In the task
 body, we implement this entry using :ada:`accept Start`. When task :ada:`T`
-reaches this point, it waits for the master task. This synchronization
-occurs in the :ada:`T.Start` statement. After the synchronization completes,
+reaches this point, it waits for some other task to call its entry.
+This synchronization
+occurs in the :ada:`T.Start` statement. After the rendezvous completes,
 the main task and task :ada:`T` again run concurrently until they synchronize
-one final time when the main task finishes.
+one final time when the main subprogram Show_Rendezvous finishes.
```
gusthoff, line 259:
```suggestion
one final time when the main subprogram :ada:`Show_Rendezvous` finishes.
```

content/courses/intro-to-ada/chapters/tasking.rst
```diff
@@ -335,9 +342,11 @@ observations:
     - Because task :ada:`T` contains an infinite loop, it always accepts calls
       to the :ada:`Reset` and :ada:`Increment` entries.
 
-    - When the main task finishes, it checks the status of the :ada:`T`
+    - When the master construct of the subtask (the Show_Rendezvous_Loop
```
gusthoff, line 345:
```suggestion
    - When the master construct of the subtask (the :ada:`Show_Rendezvous_Loop`
```

content/courses/intro-to-ada/chapters/tasking.rst
```diff
@@ -515,11 +524,11 @@ in a coordinated way, we use *protected objects*.
...
 of *protected operations*, which may be subprograms or protected
 entries. Using protected objects ensures that data is not corrupted by race
-conditions or other simultaneous access.
+conditions or other concurrent access.
 
 .. admonition:: Important
 
-    Protected objects can be implemented using Ada tasks. In fact, this was
+    Objects can be protected from concurrent access using Ada tasks. In fact, this was
     the *only* possible way of implementing them in Ada 83 (the first
```
gusthoff, line 532: Because of the previous change, we should now replace "them" with "protected objects". Otherwise, it might get confusing.

content/courses/intro-to-ada/chapters/tasking.rst
```diff
@@ -778,14 +787,14 @@ We can reuse a previous example and rewrite it to use a protected type:
 
     procedure Show_Protected_Object_Type is
 
-       protected type Obj_Type is
+       protected type PObj_Type is
```
gusthoff, line 790:
```suggestion
       protected type P_Obj_Type is
```

gusthoff (changes requested):
Thanks for reviewing this chapter. I only have a couple of comments.

content/courses/intro-to-ada/chapters/tasking.rst
sttaft: I don't believe we want to use the term "protected object" unless it is in fact an object whose type is a "protected type."  I intentionally used the term "object protected from concurrent access" to avoid the term "protected object" which has a very specific meaning in Ada.

content/courses/intro-to-ada/chapters/tasking.rst
sttaft: The term "construct" is very general.  This is probably not the place to introduce it.  I would suggest you introduce it early on in your Introduction to Ada. It is defined in 1.1.4 of the (Annotated) Ada Reference Manual:

> 15
> A syntactic category is a nonterminal in the grammar defined in BNF under “Syntax”. ...
> 16
> A construct is a piece of text (explicit or implicit) that is an instance of a syntactic category defined under “Syntax”. 

> ...
> 16.b
> Ramification: For example, an [expression](http://www.ada-auth.org/standards/2xaarm/html/AA-4-4.html#S0132) is a construct. A declaration is a construct, whereas the thing declared by a declaration is an “entity”. 
> ...

> 17
> A constituent of a construct is the construct itself, or any construct appearing within it.

content/courses/intro-to-ada/chapters/tasking.rst
sttaft: I originally said "... synchronize with the task with the entry" but that seemed quite redundant.  Feel free to rephrase it if you can find a reasonable way to do so.

content/courses/intro-to-ada/chapters/tasking.rst
sttaft: And feel free to break the long sentence into two, to avoid the multiple "which"es.

content/courses/intro-to-ada/chapters/tasking.rst
gusthoff: In this case, I prefer to be redundant. Also, the paragraph starts with "Specifically," so we can be more specific here...

content/courses/intro-to-ada/chapters/tasking.rst
gusthoff: The term "construct" is already being used in the course (and more often than I expected), so it won't change much if we use it here.

content/courses/intro-to-ada/chapters/tasking.rst
gusthoff: Actually, this admonition is in fact referring to protected objects. Maybe we could say:

"We can implement the same functionality of protected objects using just Ada tasks. In fact, this was [...]"

content/courses/intro-to-ada/chapters/tasking.rst
```diff
@@ -195,27 +200,28 @@ In this example, we're making the task :ada:`T` wait one second after each
...
+- The task with the entry waits at that point (in the :ada:`accept` statement),
   ready to accept a call to the corresponding entry from the master task.
 
-- The master task calls the task entry, in a manner similar to a procedure
-  call, to synchronize with the subtask.
+- The other task calls the task entry, in a manner similar to a procedure
+  call, to synchronize with the entry.
 
-This synchronization between tasks is called *rendez-vous*. Let's see an
+This synchronization between tasks is called *rendezvous*. Let's see an
```
sttaft, line 224: ... is called _a_ rendezvous.

content/courses/intro-to-ada/chapters/tasking.rst
sttaft: I would suggest: "In fact, this was the only way of protecting objects from concurrent access in Ada 83 (the first ..."

sttaft (commented):
Looks very good.  I still had a few comments, many of which are editorial.

content/courses/intro-to-ada/chapters/tasking.rst
gusthoff: Perfect, I've just changed as suggested.

gusthoff (approved):

PR #825, opened 2022-12-23
Adding section on the universal text buffer

gusthoff:


content/courses/advanced-ada/chapters/strings.rst
```diff
@@ -1688,11 +1688,134 @@ then the actual information for the derived type is displayed.
...
+this type and additional procedures associated with this type.
+
+.. note::
+
+   This feature was introduced in Ada 2022.
+
+Overview
+~~~~~~~~
+
+We use the :ada:`Root_Buffer_Type'Class` type to implement an universal text
```
pat-rogers, line 1704: "an universal" should be "a universal"

content/courses/advanced-ada/chapters/strings.rst
```diff
@@ -1688,11 +1688,134 @@ then the actual information for the derived type is displayed.
...
+       V := (F => 100.2,
+             I => 100);
+       Put_Line ("V = "
+                 & V'Image);
+    end Show_Put_Image;
 
+In the body of the :ada:`Put_Float_Integer` procedure, we're using the
+:ada:`New_Line`, :ada:`Increase_Indent` and :ada:`Decrease_Indent` procedures
+to improve the format of the string returned by the :ada:`Float_Integer'Image`
+attribute. Using these procedures, you can write any kind of format for your
```
pat-rogers, line 1820: In "you can write any kind of format" the word "write" seems a little awkward because the format is not what is written. Rather, the format is how the value is presented when written, or the layout used when written. So maybe it could be "you can create any kind of output format" ?

content/courses/advanced-ada/chapters/strings.rst
gusthoff: It's probably not the first time I do this mistake. ;-) Thanks for catching it!

content/courses/advanced-ada/chapters/strings.rst
gusthoff: You're totally right! I just changed it and took your suggestion.

pat-rogers (approved):

PR #842, opened 2023-02-07
Adding section on record aggregates

gusthoff:


content/courses/advanced-ada/chapters/aggregates.rst
```diff
@@ -8,11 +8,465 @@ Aggregates
...
+             Put_Line ("   Z => "
+                       & Integer'Image (P.Z3)
+                       & ")");
+          end case;
+       end Display;
+
+    end Points;
+
+To write aggregates of the :ada:`Point` type, we have to specify the :ada:`D`
+discriminant as a component of the aggregate. Also, this component must be the
```
pat-rogers, line 437: After the sentence beginning with "To write aggregates" let me suggest another sentence (before the one starting with "Also, "):
The discriminant must be included in the aggregate |mdash| and must be static |mdash| because the compiler must be able to examine the aggregate to determine if it is both complete and consistent. All components must be accounted for one way or another, as usual. But in addition, references to those components whose existence depends on the discriminant's values must be consistent with the actual discriminant value used in the aggregate. For example, for type :ada:`Point`, an aggregate can only reference the :ada:`X3`, :ada:`Y3`, and :ada:`Z3` components when :ada:`Dim_3` is specified for the discriminant :ada:`D`, otherwise those three components don't exist in that aggregate.

content/courses/advanced-ada/chapters/aggregates.rst
```diff
@@ -8,11 +8,465 @@ Aggregates
...
 
     - :arm:`4.3.1 Record Aggregates <4-3-1>`
 
 
+:ada:`<>`
+~~~~~~~~~
+
+We can use the :ada:`<>` syntax to tell the compiler to use the default value
+for specific components. However, if there's no default value for specific
+components, that component isn't initialized. For example:
```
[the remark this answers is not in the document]
gusthoff: Yes, good point, thanks!

content/courses/advanced-ada/chapters/aggregates.rst
gusthoff: Very good description, thanks a lot! I've included it (almost) verbatim.

pat-rogers (approved):

PR #846, opened 2023-02-21
Adding subsections on the Machine and Model attributes

gusthoff:


pat-rogers (commented):
This topic is a definite weakness of mine -- you may want to get another reviewer who knows this subject well.

content/courses/advanced-ada/chapters/numerics.rst
```diff
@@ -1061,26 +1061,114 @@ Let's see some examples:
...
+
+       V := 1.0E+15;
+       Put ("1.0E+15 = ");
+       F_IO.Put (Item => V);
+       New_Line;
+
+    end Show_Float_Value;
+
+If we run this example on a typical PC, we see that the value 1000000000000000.0
+was modified to 999999986991000.0. This is because 1.0 x 10\ :sup:`15` isn't
```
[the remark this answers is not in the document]
gusthoff: Yes, you're totally right: the way I phrased make it sound like the call to `Put` modified the value.

content/courses/advanced-ada/chapters/numerics.rst
```diff
@@ -1061,26 +1061,115 @@ Let's see some examples:
...
+       F_IO.Default_Exp  := 0;
+
+       Put_Line ("Original value: 1000000000000000.0");
+
+       V := 1.0E+15;
+       Put ("Machine value:  ");
+       F_IO.Put (Item => V);
+       New_Line;
+
+       V := 1.0E+15 - Float'Machine (1.0E+15);
```
pat-rogers, line 1142: It's good that you've asked Tuck to review this section. I would have bet that V was assigned zero here!

content/courses/advanced-ada/chapters/numerics.rst
```diff
@@ -1182,20 +1271,60 @@ Let's see some examples:
                  Long_Long_Float'Model_Small'Image);
     end Show_Model_Epsilon_Small;
 
-.. todo::
+Attribute: :ada:`'Model`
+^^^^^^^^^^^^^^^^^^^^^^^^
 
-    Add discussion about :ada:`'Model`.
+The :ada:`'Model` attribute is similar to the :ada:`'Machine` attribute that
```
sttaft, line 1277: There is a paragraph in AARM (G.2.1(3.b/5)) which gives some overview of the purpose of Model numbers.  I think it would be helpful to repeat it here, perhaps slimmed down a bit.  Here it is:

> In hardware that is free of arithmetic anomalies, T'Model_Mantissa, T'Model_Emin, T'Safe_First, and T'Safe_Last will yield the same values as T'Machine_Mantissa, T'Machine_Emin, T'Base'First, and T'Base'Last, respectively, and the model numbers in the safe range of the type T will coincide with the machine numbers of the type T. In less perfect hardware, it is not possible for the model-oriented attributes to have these optimal values, since the hardware, by definition, and therefore the implementation, cannot conform to the stringencies of the resulting model; in this case, the values yielded by the model-oriented parameters have to be made more conservative (that is i.e., have to be penalized), with the result that the model numbers are more widely separated than the machine numbers, and the safe range is a subrange of the base range. The implementation will then be able to conform to the requirements of the weaker model defined by the sparser set of model numbers and the smaller safe range. 

Even before this, you might want to steal something from G.2.1(1) which says:

> In the strict mode, the predefined operations of a floating point type shall satisfy the accuracy requirements specified here and shall avoid or signal overflow in the situations described. This behavior is presented in terms of a model of floating point arithmetic that builds on the concept of the canonical form (see [A.5.3](http://www.ada-auth.org/standards/22aarm/html/AA-A-5-3.html)).

The notion of "machine numbers" is pretty easy to grasp.  But without some sort of introduction, the notion of "model numbers" will probably be pretty mysterious.

Note that one place where the model and the machine are most likely to differ is in "denormalized" numbers -- numbers very close to zero, where the machine numbers begin to provide fewer bits of precision than normal, while the model numbers hit the Model_Emin, and cannot represent denormalized numbers at all. You might want to have an example that illustrates the difference with such "teeny" numbers.

content/courses/advanced-ada/chapters/numerics.rst
```diff
@@ -1182,20 +1271,60 @@ Let's see some examples:
...
+    procedure Show_Model_Attribute is
+       package F_IO is new Ada.Text_IO.Float_IO (Float);
+
+       V : Float;
+    begin
+       F_IO.Default_Fore := 3;
+       F_IO.Default_Aft  := 1;
+       F_IO.Default_Exp  := 0;
+
+       Put_Line ("Original value: 1000000000000000.0");
```
sttaft, line 1304: It would be easier to read if you put '_' every three digits.

gusthoff: Yes, you're right. I changed it here and in other places.

content/courses/advanced-ada/chapters/numerics.rst
gusthoff: I agree with your suggestions, and I'll work on a better introduction to model numbers in a separate pull request.

PR #861, opened 2023-04-08
Adding section on type conversion

gusthoff:


content/courses/advanced-ada/chapters/types.rst
```diff
@@ -1643,19 +1643,913 @@ want to declare objects of this type.
...
+       Put_Line ("AI_D: "
+                 & AI_D'Image);
+    end Show_Conversion;
+
+Note that both arrays must have the same number of components in order for the
+conversion to be successful. (Sliding is fine, though.) In this example, both
+arrays have the same range: :ada:`Common_Range`.
+
+We can also convert between array types that aren't derived one from the
+other. As long as the components are of the same type, the conversion between
```
pat-rogers, line 1941: The index (sub)types are part of the requirements too. Overall:
Component types must be the same type
Index types must be the same or convertible
Dimensionality must be the same
Bounds must be compatible (not necessarily equal)

content/courses/advanced-ada/chapters/types.rst
```diff
@@ -1643,19 +1643,913 @@ want to declare objects of this type.
...
+types. However, we can still convert between them in the same way as we did in
+the previous code examples.
+
+
+View conversion
+~~~~~~~~~~~~~~~
+
+As mentioned before, view conversions just allow names to be converted. Thus,
+we cannot use expressions in this case.
+
```
pat-rogers, line 2086: I think it is important to relate the concept of "view" and then, in that context, say that the conversion changes the view of a given object to the target type's view, both in terms of components that exist and operations that are available. It doesn't really change anything at all in the value itself.

content/courses/advanced-ada/chapters/types.rst
gusthoff: Yes, you're right. (I added these requirements to the paragraph.)

content/courses/advanced-ada/chapters/types.rst
gusthoff: Yes, I agree. (I hope the explanation that I've just added is clear enough.)

pat-rogers (approved):

PR #923, opened 2023-08-19
Adding section on design strategies for access types

gusthoff:


content/courses/advanced-ada/parts/resource_management/access_types.rst
```diff
@@ -3445,26 +3445,239 @@ caller.
...
+       Destroy (Obj_1);
+       Destroy (Obj_2);
+    end Show_Access_Type_Abstraction;
+
+In this example, we hide an access type in the :ada:`Info` type |mdash| a
+limited private type. We allocate an object of this type in the :ada:`Init`
+function and deallocate it in the :ada:`Destroy` procedure. Also, we make
+sure that the reference isn't copied in the :ada:`Copy` function |mdash|
+we only copy the designated value in this function. This strategy eliminates
+the possibility of dangling references, as each reference is encapsulated in
```
pat-rogers, line 3551: I don't understand how that makes dangling references impossible. For example, given an object named X, of type Info, wouldn't a call to Destroy(X), followed by a call to To_Integer(X), dereference a null access value, ie a dangling reference?

gusthoff: Yes, you're right: this code example was bad. I've just reworked it and improved it a bit.

content/courses/advanced-ada/parts/resource_management/access_types.rst
```diff
@@ -3445,26 +3445,401 @@ caller.
...
+:ada:`Limited_Controlled` type from the :ada:`Ada.Finalization` package,
+we had to adapt the prototype of the subprograms from the
+:ada:`Access_Type_Abstraction`. In this version of the code, we only have
+the allocation taking place in the :ada:`To_Info` procedure, but we don't have
+a :ada:`Destroy` procedure for deallocation: this call was moved to the
+:ada:`Finalize` procedure.
+
+Since objects of the :ada:`Info` type |mdash| such as :ada:`Obj_1` in the
+:ada:`Show_Access_Type_Abstraction` procedure |mdash| are now controlled, the
+:ada:`Finalize` procedure is automatically called when they get out of scope.
```
pat-rogers, line 3833: "get out of scope" might be more commonly written "go out of scope"

content/courses/advanced-ada/parts/resource_management/access_types.rst
```diff
@@ -3445,26 +3445,401 @@ caller.
...
+          Obj := New_Str;
+       end Append;
+
+       --
+       --  PRIVATE SUBPROGRAMS
+       --
+
+       procedure Initialize (Obj : in out Info) is
+       begin
+          Obj.Str_A := null;
```
pat-rogers, line 3721: Since null is the component's default value anyway, why do we need procedure Initialize?

gusthoff: You're right, this isn't strictly needed. However, I'd like to keep this line in order to illustrate the automatic calls to  `Initialize` and `Finalize` for controlled types in this context. I'll add a big note to the procedure explaining that this assignment is actually useless in "production" code. I hope this is OK with you.

gusthoff: Also, I'll add a call to `Put_Line` in the comments.

gusthoff: Anyway, if you prefer, I could replace that assignment by a `null;`.

pat-rogers: A note about it not being necessary for that specific component type would be a good idea, but otherwise I've no objection to keeping it.

pat-rogers (approved):

PR #932, opened 2023-09-29
Widget: adding help button for each compiler switch

gusthoff:


gusthoff:
Snapshot of the help button in action:

switches_menu_help_button_2

PR #997, opened 2024-02-17
Adding chapter: "Controlling Object Initialization and Creation"

gusthoff:
Chapter was written by Pat Rogers.

### DIFF
--- a/content/courses/ada-idioms/chapters/controlling_obj_initialization_creation.rst
+++ b/content/courses/ada-idioms/chapters/controlling_obj_initialization_creation.rst
@@ -10,7 +10,8 @@ Motivation
 ----------
 
 Developers are responsible for ensuring that no uninitialized objects are read
-in Ada. Default initialization is a good way to meet this requirement because
+in Ada programs. Default initialization is a good way to meet this requirement
+because
 it is guaranteed to happen and requires no actions on the part of the client
 code. But of the many kinds of types provided by Ada, only access types have a
 language-defined default initial value. Fortunately, Ada supports user-defined
--- a/content/courses/ada-idioms/chapters/controlling_obj_initialization_creation.rst
+++ b/content/courses/ada-idioms/chapters/controlling_obj_initialization_creation.rst
@@ -27,7 +27,7 @@ scalar components.
 
 These facilities are often sufficient to express an abstraction's initial
 state. For example, we can expect that container objects will be initially
-empty. For example, a bounded stack ADT is almost certainly represented by a
+empty. Consider a bounded stack ADT. The representation is likely a
 record type containing an array component and a :ada:`Top` component indicating
 the index of the last array component used. We can default initialize objects
 to the empty state simply by setting :ada:`Top` to zero in the record
@@ -41,8 +41,8 @@ component's declaration:
        Top    : Natural := 0;
     end record;
 
-For an unbounded container such as a simple binary tree, we'll use an access
-type for the representation, and the automatic default value null initializes
+For an unbounded container such as a simple binary tree, if the representation
+is an access type, the automatic default value :ada:`null` initializes
 :ada:`Tree` objects to the empty state.
 
 .. code-block:: ada
@@ -84,7 +84,8 @@ function call.
        ...
     end;
 
-The initial value can also specify constraints, if required. The object
+The initial value can also specify constraints, if required.
+In the code above, the object
 :ada:`Prompt` has a lower bound of :ada:`Positive'First` and an upper bound set
 to the length of the literal. The specific bounds of :ada:`Reply` are
 determined by the function, and need not start at :ada:`Positive'First`.
@@ -281,7 +282,8 @@ The parentheses are required as usual, but the *box* symbol appears inside,
 instead of one or more discriminant declarations. The box symbol always
 indicates *not specified here* so in this case no discriminants are included in
 the view. There may or may not be discriminants in the full view, but client's
-don't have compile-time visibility to that information.
+don't have compile-time visibility to that information because the type is
+private.
 
 Unknown discriminants can be specified for various kinds of types, not only
 private types. See the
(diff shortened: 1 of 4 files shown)


content/courses/ada-idioms/chapters/controlling_obj_initialization_creation.rst
```diff
@@ -0,0 +1,1177 @@
...
+       Put_Line (Obj1'Image);
+       Put_Line (Obj2'Image);
+       Put_Line ("Done");
+    end Main;
+
+In the above, :ada:`Assertion_Error` is raised by the elaboration of
+:ada:`Obj2` because the DIC check necessarily fails. There is no check on the
+declaration of :ada:`Obj1` because it is initialized, explicitly.
+
+
```
pat-rogers, line 711: Let's add the following here:

To recap, we can ensure initialization for objects of the type by detecting, during elaboration at run-time, any objects not explicitly initialized.

This approach is sufficient because when elaboration of an object declaration raises an exception, no use of that object is possible. That's guaranteed because the frame containing that declarative part is immediately abandoned and the exception is propagated up to the previous level. A local handler never can apply. But even if there is a matching handler in the previous level, there's really nothing much to be done. Re-entering the frame containing the declaration will raise the exception all over again, necessarily. Thus the code will have to be changed and recompiled, meeting the goal of the idiom.

We can illustrate this assurance using Storage_Error. Consider the following program, in which the main procedure calls an inner procedure P:

with Text_IO;  use Text_IO;

procedure Main is

   procedure P (Output : out Float) is
      N : array (Positive) of Float;   -- Storage_Error is likely
   begin
      Put_Line ("P's body assigns N's components and uses them");
      --  The following indexes and component values are arbitrary 
      --  and used purely for illustration...
      N := (others => 0.0);  
      --  other computations and assignments to N ...      
      Output := N (5);  
   exception
      when Storage_Error =>
         Output := N (1);
   end P;

   X : Float;
   
begin
   P (X);
   Put_Line (X'Image);   
   Put_Line ("Done");
exception
   when Storage_Error =>
      Put_Line ("Main completes abnormally");
end Main;

When Main calls P, the elaboration of the declarative part of P almost certainly fails because there is insufficient storage to allocate to the object P.N, hence Storage_Error is raised. (If your machine can handle the above, congratulations.) Even though procedure P has a handler specifically for Storage_Error, that  handler never applies because the declarative part is immediately abandoned. Instead, the exception is raised in the caller, where it can be caught. This behavior is essential to ensure that problematic objects are not referenced in the local handlers. In the above, the handler in P for Storage_Error references the object P.N to assign the P.Output parameter. If that assignment could happen -- again, it cannot -- what would it mean, functionally? No one knows.

Handling Storage_Error is a little tricky anyway. Does the OS give the program a chance to execute a handler? If so, is there sufficient storage remaining to execute the exception handler's statements? In any case you can see the problem that the declaration failure semantics preclude.

Therefore, although the DIC solution is not enforced at compile-time, it is nevertheless sufficient to ensure no uninitialized object of the type can be used.

content/courses/ada-idioms/chapters/controlling_obj_initialization_creation.rst
```diff
@@ -0,0 +1,1177 @@
...
+functions. Doing so removes any means for initializing objects of the type, and
+since the type is indefinite there is then no way for clients to declare
+objects at all. The compiler again enforces this solution.
+
+For a concrete example, we can apply the Singleton design pattern to represent
+the *time stamp counter* (:wikipedia:`TSC <Time_Stamp_Counter>`) provided by
+x86 architectures. The TSC is a 64-bit hardware register incremented once per
+clock cycle, starting from zero at power-up. We can use it to make a timestamp
+abstraction. As explained on that
+:wikipedia:`Wikipedia page <Time_Stamp_Counter>`, some care is required when
```
pat-rogers, line 727: "As explained on that" => "As explained by"

content/courses/ada-idioms/chapters/controlling_obj_initialization_creation.rst
```diff
@@ -0,0 +1,1177 @@
...
+
+    procedure Demo_TimeStamp is
+    begin
+       for K in 1 .. 10 loop
+          Put_Line (Timestamp.Counter.Sample'Image);
+       end loop;
+    end Demo_TimeStamp;
+
+The above calls the :ada:`Timestamp.Counter` function and then implicitly
+dereferenced the resulting access value to call the :ada:`Sample` function
```
pat-rogers, line 831: dereferenced  => dereferences

content/courses/ada-idioms/chapters/controlling_obj_initialization_creation.rst
```diff
@@ -0,0 +1,1177 @@
...
+using the distinguished receiver syntax. The resulting number is then converted
+to a :ada:`String` value and output to standard output.
+
```
pat-rogers, line 834: standard output => Standard_Output

content/courses/ada-idioms/chapters/controlling_obj_initialization_creation.rst
```diff
@@ -0,0 +1,1177 @@
...
+Function :ada:`Counter` creates the single object that this singleton
+implementation creates. It does so by lazily allocating an object dynamically.
+If :ada:`Counter` is never called (because some subclass is used instead) then
+no object of type :ada:`Cycle_Counter` is created. At most one
+:ada:`Cycle_Counter` object is ever created.
+
+We could instead declare :ada:`The_Instance` as a :ada:`Cycle_Counter` object
+in the package body, mark it as aliased, and return a corresponding access
+value designating it. But when objects are large, declaring one that is might
```
pat-rogers, line 878: "that is might" => "that might"

pat-rogers (approved):
Thanks!

PR #998, opened 2024-02-18
Various refinements

pat-rogers:
Change the title and motivation summary for the constructors idiom to say it is about ADTs, not just type extensions.
Refine the component access idiom to give example for immutably limited types, etc.


gusthoff:
> Change the title and motivation summary for the constructors idiom to say it is about ADTs, not just type extensions. Refine the component access idiom to give example for immutably limited types, etc.

Since you're changing the chapter title, we should also rename the file (i.e. the actual .rst file for that chapter) accordingly. I'll take care of this separately later on.

gusthoff (approved):

gusthoff:
Done: #1000

gusthoff:
> Since you're changing the chapter title, we should also rename the file (i.e. the actual .rst file for that chapter) accordingly. I'll take care of this separately later on.

Done: #1000

pat-rogers:
Thanks!

PR #1012, opened 2024-03-09
Adding section on initialization of controlled types

gusthoff:


content/courses/advanced-ada/parts/resource_management/controlled_types.rst
```diff
@@ -569,16 +569,513 @@ Finally, the new version of object :ada:`A` (the one whose ID is 3) and object
...
+
+    end Workers;
+
+    with Ada.Text_IO; use Ada.Text_IO;
+
+    package body Workers is
+
+       task body Worker is
+       begin
+          Put_Line ("Activating Worker task...");
```
pat-rogers, line 971: You might want to move this Put_Line call to a function called within the declarative part, since activation of a given task is the elaboration of the task object's declarative part. But that might be more precision than necessary...

gusthoff: On the contrary: I think this is an excellent idea! I've just changed the code accordingly. Thanks for the suggestion! 

Also, I took the opportunity to add some diagrams with the calling sequence.

pat-rogers (approved):

PR #1015, opened 2024-03-19
Use a more compact layout for the front page ebook covers

Fabien-Chouteau:
The current layout is too big for the growing number of books on Learn, the latest books are too far down the page. The goal of this patch is to have more e-books (covers and download buttons) displayed on the screen using a compact layout.

### DIFF
--- a/content/index.rst
+++ b/content/index.rst
@@ -120,233 +120,214 @@
 
     .. container:: frontpage-ebooks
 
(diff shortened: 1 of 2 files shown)


Fabien-Chouteau:
Here's what it looks like:
2024-03-19-160819_3840x1200_scrot

PR #1018, opened 2024-03-23
Adding section on assignment of controlled types

gusthoff:


content/courses/advanced-ada/parts/resource_management/controlled_types.rst
```diff
@@ -1189,13 +1189,311 @@ This diagram shows the initialization sequence:
...
 
     - :arm22:`Assignment and Finalization <7-6>`
 
-.. todo::
 
-    Complete section!
+Assignment using anonymous object
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The :arm22:`Ada Reference Manual <7-6>` mentions that an anonymous object is
```
pat-rogers, line 1205: I don't know where I would suggest saying this, but I suggest that the text says something about the fact that an actual assignment is a full bit-wise copy of the entire right-hand side to the entire left-hand side, and that adjustment is a way to "work around" that, when necessary. You may have already said that elsewhere, but in that case I think it's worth repeating here. Otherwise the need for Adjust will be lost to readers. (And apologies if you've already said this elsewhere and consider that to be sufficient, in which case feel free to ignore this!)

content/courses/advanced-ada/parts/resource_management/controlled_types.rst
```diff
@@ -1189,13 +1189,311 @@ This diagram shows the initialization sequence:
...
+    end Simple_Controlled_Types;
+
+    with Ada.Text_IO; use Ada.Text_IO;
+
+    package body Simple_Controlled_Types is
+
+       function Default_Init return Integer is
+       begin
+          Put_Line ("Default_Init: Integer...");
+          return 42;
```
pat-rogers, line 1429: 42 is, of course the only appropriate value! :-)

content/courses/advanced-ada/parts/resource_management/controlled_types.rst
```diff
@@ -1189,13 +1189,311 @@ This diagram shows the initialization sequence:
...
+
+       Put_Line ("----------");
+       Put_Line ("A := B");
+       A := B;
+       Put_Line ("----------");
+    end Show_Controlled_Types;
+
+When running this code, we see that the :ada:`S1` and :ada:`S2` components are
+adjusted before the adjustment of the parent type :ada:`T` takes place.
+
```
pat-rogers, line 1476: Consider saying something to the effect that the components must be done before the enclosing object because the semantics of the adjustment of the whole might depend on the states of the parts (the components) so those states must already be in place.

content/courses/advanced-ada/parts/resource_management/controlled_types.rst
gusthoff: Yes, I totally agree! I've just added this phrase twice in the text. Also, I added a paragraph about deep copy as an example.

content/courses/advanced-ada/parts/resource_management/controlled_types.rst
gusthoff: Yes, good point! I've just added this phrase (up in the introduction of this subsection).

pat-rogers (approved):
Nice!

PR #1025, opened 2024-04-05
Use "local units" instead of "subunits" in explanation

yannickmoy:
Subunits are something different in Ada, as dedfined in Ada RM 1.3.3:
"body of a program unit that can be compiled separately from its
enclosing program unit"


yannickmoy:
@gusthoff I don't understand why I can't rebase this tiny patch on top of `AdaCore:master`. Doing it locally end up with tons of conflicts...

gusthoff:
This PR was replaced by #1032.

PR #1032, opened 2024-04-12
Use "local units" instead of "subunits" in explanation

gusthoff:


yannickmoy:
Thanks Gustavo! It was still in my list of todo, did not have the time to get a clean clone of the repo to reapply the patch.

yannickmoy (approved):

PR #1033, opened 2024-04-19
Improvements to section on per-object expressions

gusthoff:
Integrating suggestions by Tucker Taft.

### DIFF
--- a/content/courses/advanced-ada/parts/data_types/records.rst
+++ b/content/courses/advanced-ada/parts/data_types/records.rst
@@ -708,12 +708,12 @@ so-called *current instance*.
 
    - :arm22:`3.8 Record Types <3-8>`
 
-Restrictions on per-object expressions
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-There are some important restrictions on per-object expressions and per-object
-constraints:
 
+Restrictions
+~~~~~~~~~~~~
+
+There are some important restrictions on per-object constraints:
 
 #. We can only use access attributes (:ada:`T'Access` and
    :ada:`T'Unchecked_Access`) in per-object range constraints.
--- a/content/courses/advanced-ada/parts/data_types/records.rst
+++ b/content/courses/advanced-ada/parts/data_types/records.rst
@@ -709,6 +709,7 @@ so-called *current instance*.
    - :arm22:`3.8 Record Types <3-8>`
 
 
+.. _Adv_Ada_Per_Object_Expressions_Restrictions:
 
 Restrictions
 ~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


content/courses/advanced-ada/parts/data_types/records.rst
```diff
@@ -672,18 +677,149 @@ declaring. For example:
...
+
+   - :arm22:`3.8 Record Types <3-8>`
+
+Restrictions on per-object expressions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There are some important restrictions on per-object expressions and per-object
+constraints:
+
+#. From all type attributes, only access attributes (:ada:`T'Access` and
```
sttaft, line 717: This is a bit confusing.  T'Access is not actually a *type* attribute, it is an attribute of the current instance *object*.

content/courses/advanced-ada/parts/data_types/records.rst
```diff
@@ -672,18 +677,149 @@ declaring. For example:
...
+   :ada:`T'Unchecked_Access`) can be used in a per-object expression.
```
sttaft, line 718: Actually, per-object expressions don't have many limitations, it is mostly per-object constraints that have limitations.  You can refer to any defined attribute of the current instance *object* in a per-object expression used as the default expression for a component.

gusthoff: You're right, the way it's written is very confusing. I've changed this part, and also added a subsection on default values.

content/courses/advanced-ada/parts/data_types/records.rst
```diff
@@ -708,14 +708,99 @@ so-called *current instance*.
...
+
+.. _Adv_Ada_Per_Object_Expressions_Restrictions:
+
+Restrictions
+~~~~~~~~~~~~
+
+There are some important restrictions on per-object constraints:
+
+#. We can only use access attributes (:ada:`T'Access` and
+   :ada:`T'Unchecked_Access`) in per-object range constraints.
```
sttaft, line 803: This should probably say simply "in per-object constraints."  In fact, they are only usable in discriminant constraints, and as you state on the next line, per-object range constraints are never allowed.

content/courses/advanced-ada/parts/data_types/records.rst
```diff
@@ -708,14 +708,99 @@ so-called *current instance*.
...
+
+
+.. _Adv_Ada_Per_Object_Expressions_Restrictions:
+
+Restrictions
+~~~~~~~~~~~~
+
+There are some important restrictions on per-object constraints:
+
+#. We can only use access attributes (:ada:`T'Access` and
```
sttaft, line 802: I would suggest you start by saying that per-object range constraints are not allowed.
Then talk about index/discriminant constraints, where only a direct name of a discriminant is permitted.
And *then* talk about T'Access, since it is really much less common, and generally a bit obscure.

sttaft:
Looks good!

sttaft (approved):
Looks good.

PR #1076, opened 2024-07-18
Add CWE to each standard

frank-at-adacore:
Add a new subsection (like the ISO/IEC one) to list CWE's for each rule

### DIFF
--- a/content/courses/Guidelines_for_Safe_and_Secure_Ada_SPARK/chapters/guidelines/concurrency/con01_use_the_ravenscar_profile.rst
+++ b/content/courses/Guidelines_for_Safe_and_Secure_Ada_SPARK/chapters/guidelines/concurrency/con01_use_the_ravenscar_profile.rst
@@ -89,9 +89,9 @@ Applicable Vulnerability within ISO TR 24772-2
 Applicable Common Weakness Enumeration
 ++++++++++++++++++++++++++++++++++++++++
 
-* CWE-362 - Concurrent Execution using Shared Resource with Improper Synchronization
-* CWE-367 - Time-of-check Time-of-use (TOCTOU) Race Condition
-* CWE-366 - Race Condition within a Thread
+* :cwe:`CWE-362 - Concurrent Execution using Shared Resource with Improper Synchronization <362>`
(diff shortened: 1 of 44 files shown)


content/courses/Guidelines_for_Safe_and_Secure_Ada_SPARK/chapters/guidelines/concurrency/con01_use_the_ravenscar_profile.rst
```diff
@@ -85,6 +85,14 @@ Applicable Vulnerability within ISO TR 24772-2
 * 6.62 Concurrency - Premature termination [CGS]
 * 6.63 Lock protocol errors [CGM]
 
+++++++++++++++++++++++++++++++++++++++++
+Applicable Common Weakness Enumeration
+++++++++++++++++++++++++++++++++++++++++
+
+* CWE-362 - Concurrent Execution using Shared Resource with Improper Synchronization
```
gusthoff, line 92: Since the cwe.mitre.org website has a well-formatted URL for the CWEs, I´d strongly suggest that we create a Sphinx "extlink" for it, so that readers can go to the corresponding page instead of having to "google" for each of them. This is how it would look like: 

```
:cwe:`CWE-362 - Concurrent Execution using Shared Resource with Improper Synchronization <362>`
```

Also, I volunteer to make the appropriate changes. :-)

frank-at-adacore: I appreciate you volunteering :)
The link is actually in the "References" section

gusthoff: Yes, I saw that link, but the current version still forces readers to lookup for specific CWEs on that website, which in turn requires extra typing and multiple clicks. If they had an hyperlink here, it would be just a single click.

gusthoff (approved):

PR #1112, opened 2024-09-14
Avoid bash error on undefined variable `v`

reznikmm:


gusthoff (approved):
Thanks a lot for catching and fixing this issue!

Issue #1101, opened 2024-09-04
Multiple file tabbed editor view lacks UI affordances, leading to confusion


brgaeke:
I witnessed the following with a student: in the cases where there are multiple files available in the tabbed editor view, the non-selected tabs (which appear to the right of the initially selected tab) do not bear any clear indication that they can be clicked on.  This leads to students potentially not understanding what they are supposed to edit (e.g., the student tries to put the body into classify_number.ads instead of classify_number.adb)

![image](https://github.com/user-attachments/assets/e1f13b8b-e462-4fee-b2b8-edddfe4ccc29)

Red arrow shows the UI elements I'm talking about.

This might be fixed by
1. modifying the UI elements so they look more selectable (compare against Github's "new issue" interface, for example)
2. putting some text in the lab instructions instructing students that they can, and are expected to, click on the other files to edit them
3. selecting by default the file that is more of interest to completing the lab (e.g., the body instead of the spec)
4. adding some kind of mouse-over effect to the tabs

gusthoff:
Hey @brgaeke,

You're right: it's not obvious that those are clickable tabs. We will work on improving the layout. Thanks for the suggestions!

gusthoff:
* Design of the source-code tabs of the widget was improved (#1114).
* Instructions about source-code tabs were added to the labs (#1115).

PR #1114, opened 2024-09-27
Widget: improvements to source-code tabs

gusthoff:
Fix some issues reported in #1101


gusthoff:
* Adding (dotted) border to each source-code tab.
* Using visible mouse-over effect for the source-code tabs.

2024-09-27_learn_tabbed-src

PR #1145, opened 2024-11-21
changes made in response to Richard's first set of comments

pat-rogers:


content/courses/ada-idioms/chapters/abstract_data_types.rst
```diff
@@ -331,13 +338,14 @@ change in client usage, such as performance changes, but it will not be a
...
-fundamental to creating abstractions in class-oriented languages. Not every
-type can be private, as illustrated by the client expectation for array
-indexing in Ada. Not every type should be private, for example those that are
-explicitly numeric. But the ADT should be the default design idiom when
-decomposing a problem into a solution.
+The private type is the fundamental approach to creating abstractions in
+Ada, just as the use of the *public*, *private*, and *protected* parts of
+classes is fundamental to creating abstractions in class-oriented
+languages. Not every type can be private, as illustrated by the client
+expectation for array indexing in Ada prior to Ada 2012. Not every type
```
gusthoff, line 345: Adding "prior to Ada 2012" here makes the uninformed reader wonder why is it now different in Ada 2012... Maybe we should come up with better examples of the necessity of non-private types.

content/courses/ada-idioms/chapters/constructor_functions_for_abstract_data_types.rst
```diff
@@ -42,24 +47,54 @@ extension:
...
+Primitive operations, and only primitive operations, are inherited during
+type derivation.
+
+If you think in terms of Abstract Data Types all these rules make sense.
 
-Therefore, Ada language rules prevent constructor functions from being legally
-inherited, even though they are primitive operations for the type.
+Now, here's the rub.
+
+Constructor functions require that same compile-time visibility so the
```
gusthoff, line 90:
```suggestion
Now, here's the problem: 
constructor functions require that same compile-time visibility so the
```

gusthoff (approved):

PR #1146, opened 2024-11-24
Adding subsection on Full_Access_Only aspect

gusthoff:


content/courses/advanced-ada/parts/data_types/types_representation.rst
```diff
@@ -2775,6 +2775,238 @@ these array declarations are equivalent:
     end Shared_Var_Types;
 
 
+:ada:`Full_Access_Only`
+~~~~~~~~~~~~~~~~~~~~~~~
+
+A full access object is an object that requires that read or write operations
+on this object be performed by reading or writing all bits of the object (i.e.
```
pat-rogers, line 2785: "this object be performed" is a little awkward. Maybe "this object are performed"  instead.

content/courses/advanced-ada/parts/data_types/types_representation.rst
```diff
@@ -2775,6 +2775,238 @@ these array declarations are equivalent:
...
+specific days. In the :ada:`Show_Alarm` procedure, we set the alarm to just
+ring on Mondays and Wednesdays.
+
+The compiler may generate multiple operations for the update of the
+:ada:`A.Days (Mon)` and :ada:`A.Days (Wed)` components. If we use
+:ada:`Full_Access_Only` aspect, we may combine those updates in a single
+operation |mdash| in addition, we have to use a
+:ref:`delta aggregate <Adv_Ada_Delta_Aggregates>`.
+
+Note that we haven't discuss the topic of delta aggregates yet: we'll do that
```
pat-rogers, line 2919: "discuss" => "discussed"

content/courses/advanced-ada/parts/data_types/types_representation.rst
```diff
@@ -2775,6 +2775,238 @@ these array declarations are equivalent:
     end Shared_Var_Types;
 
 
+:ada:`Full_Access_Only`
+~~~~~~~~~~~~~~~~~~~~~~~
+
+A full access object is an object that requires that read or write operations
+on this object be performed by reading or writing all bits of the object (i.e.
+the *full object*) at once. Accordingly, a full access type is a type whose
```
pat-rogers, line 2786: "at once."
I think this statement doesn't go far enough to distinguish such objects (and types) from atomic objects. The resulting dynamics semantics of these objects is that, even when the source code only reads or updates a part of the object, the entire object is read or updated (and the unchanged bits are written back). 
I believe the examples are also missing the above point.
See AARM22 C.6 {19.1/5 and 19.C/5}.
For examples, I can point you to some of the STM32 Ada Driver Library drivers, which lead to the request for this aspect. In these cases the STM device hardware reference manual requires volatile full access even when reading/updating subcomponents of the memory-mapped device.
The effect for an update is the same as if the source code does a read of the entire object into a temporary object, makes local subcomponent changes to that temporary, and then writes the temporary value back to the original object. For a read, the entire object is read into the temporary object, and then parts of the temporary are accessed individually. That's what the actual C version of the drivers actually does.  We didn't want to do that in Ada at the source code level so we have the compiler do it.

gusthoff: Thanks for the clarification! I'll look into the Ada Driver Library for inspiration and rework the source-code examples later on.

content/courses/advanced-ada/parts/data_types/types_representation.rst
gusthoff: Thanks for the correction!

PR #1172, opened 2025-01-16
Fix for null access support in example

leogermond:
Fixes null access exceptions for example in advanced Ada / Resource Management / Access Types / Access Type Abstraction example when calling `Copy` (both versions) with a `Info = null`


gusthoff (approved):
Thanks for the fix!

PR #1184, opened 2025-02-14
Fix typo in ada-in-practice/introduction.rst

PaulCapron:
A ‘**r**’ was missing in “[Be**r**trand Meyer](https://en.wikipedia.org/wiki/Bertrand_Meyer)”.


gusthoff:
Thanks for the correction!

PR #1245, opened 2025-07-06
Adding section on discrete numeric types

gusthoff:


content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -1329,22 +1329,457 @@ input parameter :ada:`V` (of :ada:`One_To_Ten'Base` type) in the runtime output
...
+
+.. _Adv_Ada_Integer_Types:
+
+Integer types
+~~~~~~~~~~~~~
+
+In the :ref:`Introduction to Ada <Intro_Ada_Integers>` course, we mentioned
+that you can define your own integer types in Ada. In fact, you're almost
+expected to do so, as Ada only guarantees the existence of a single integer
+type |mdash| and a few optional integer types. Even though a specific compiler
```
pat-rogers, line 1367: Maybe "type |mdash| and a few optional integer types." should be
"type |mdash| and the names of a few optional integer types."  ??

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -1329,22 +1329,457 @@ input parameter :ada:`V` (of :ada:`One_To_Ten'Base` type) in the runtime output
...
+:ref:`signed integer types <Adv_Ada_Integer_Types>`, and then move on to
+:ref:`modular types <Adv_Ada_Modular_Types>`.
+
+.. _Adv_Ada_Integer_Types:
+
+Integer types
+~~~~~~~~~~~~~
+
+In the :ref:`Introduction to Ada <Intro_Ada_Integers>` course, we mentioned
+that you can define your own integer types in Ada. In fact, you're almost
```
pat-rogers, line 1365: Consider changing "you're almost" here to "typically you're"

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -1329,22 +1329,457 @@ input parameter :ada:`V` (of :ada:`One_To_Ten'Base` type) in the runtime output
...
+                 & " bits");
+       Put_Line ("Coefficient'First :"
+                 & Coefficient'First'Image);
+       Put_Line ("Coefficient'Last  :"
+                 & Coefficient'Last'Image);
+    end Show_Custom_Integer_Types;
+
+In this example, we declare the 32-bit :ada:`Coefficient` type. We ensure that
+it's a 32-bit type by explicitly writing :ada:`range -2**31 .. +2**31 - 1`.
+
```
pat-rogers, line 1603: And if the specified range cannot be supported by the target machine, the Ada compiler will reject the source code containing the type declaration (and all clients of that code). It is important (you may have done so elsewhere) to explain that this is a different approach to portability than that of, say, C, where for example type int is always defined and hence the client code always compiles, but won't necessarily work at run-time. In that case at best you find the problem during testing, which is comparatively expensive. Worse, if you don't find out until after deployment, the cost to fix it is much, much higher. In contrast, with a user-specified integer type, if the specified range cannot be supported by the (perhaps new) target machine, you find out at compile-time, which is far less expensive and more robust too.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -1329,22 +1329,457 @@ input parameter :ada:`V` (of :ada:`One_To_Ten'Base` type) in the runtime output
...
+                 & Modular'Size'Image
+                 & " bits");
+       Put_Line ("Modular'First :"
+                 & Modular'First'Image);
+       Put_Line ("Modular'Last  :"
+                 & Modular'Last'Image);
+    end Show_Modular_Types;
+
+This example declares the 32-bit modular type :ada:`Modular`.
+
```
pat-rogers, line 1748: Probably should emphasize that the specified modulus need not be a power of 2. I have found that C programmers automatically assume that restriction.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -1329,22 +1329,457 @@ input parameter :ada:`V` (of :ada:`One_To_Ten'Base` type) in the runtime output
...
+.. _Adv_Ada_Custom_Integer_Types:
+
+Custom integer types
+^^^^^^^^^^^^^^^^^^^^
+
+As we've mentioned before, numeric data types such as :ada:`Integer` or
+:ada:`Long_Integer` might not provide enough control over the ranges |mdash|
+i.e. the range selected by the compiler may not correspond to the expected
+range of the numeric algorithm we're implementing. Therefore, it is best to
+simply define custom ranges. To do that, you should evaluate the algorithm and
```
pat-rogers, line 1567: I think you could simplify the above to 
"As we've mentioned before, for the language-defined numeric data types such as :ada:`Integer` or
:ada:`Long_Integer` the range selected by the compiler may not correspond to the required
range of the numeric algorithm we're implementing. Therefore, it is best to
simply declare custom types with the necessary ranges specified."

content/courses/advanced-ada/parts/data_types/numerics.rst
gusthoff: Good point! In the section I'm currently working on (about floating-point types), I've added a subsection about illegal ranges, but I forgot to discuss the topic in this section. I'll include a subsection and add your text in an admonition.

PR #1253, opened 2025-07-27
Adding section on floating-point types

gusthoff:


gusthoff:
Thanks for reviewing and all the suggestions!

PR #1255, opened 2025-08-15
Adding section on Resource Acquisition Is Initialization (RAII)

gusthoff:
Section was written by Patrick Rogers.

### DIFF
--- a/content/courses/ada-in-practice/chapters/raii.rst
+++ b/content/courses/ada-in-practice/chapters/raii.rst
@@ -32,12 +32,11 @@ contributions were reified in
 :wikipedia:`Concurrent Pascal <Concurrent_Pascal>`, a concrete programming
 language.), Ada defines a protected object (PO) as a concurrency construct that
 is higher-level and more robust than mutexes and semaphores. Those advantages
-accrue because the bodies of the protected operations implement only the
-functional requirements, with guaranteed mutually exclusive access to the
-protected data. The underlying implementation, i.e., the run-time library,
-provides the necessary exclusive access, and also thread management. As a
-result, the lock is guaranteed to be acquired, and likewise, guaranteed to be
-relinquished, even in the face of exceptions.
+accrue because the bodies of the protected operations are only responsible for
+implementing the functional requirements. The underlying run-time library is
+responsible for implementing the mutually exclusive access, and also thread
+management. As a result, the source code is much simpler and is robust even in
+the face of exceptions.
 
 However, a protected object is not always appropriate. Consider an existing
 sequential program that makes calls to visible procedures provided by a
@@ -154,7 +153,8 @@ protected procedure bodies both call a potentially blocking operation, in this
 case :ada:`Ada.Text_IO.Put_Line`. Erroneous execution is the result. It might
 work as intended when executed, or it might do something else, or, if detected,
 :ada:`Program_Error` will be raised. On a run-time library built on top of an
-operating system, it may work because the OS may provide thread locking
+operating system, it may work as intended because the OS may provide thread
+locking
 mechanisms that the run-time library can use. In that case a blocking operation
 just suspends the caller thread's execution temporarily without releasing the
 PO lock. Although the blocking operation would allow some other caller task to
@@ -172,7 +172,8 @@ only potentially blocking operation defined by the language.
 Note that moving the calls to :ada:`Put_Line` out of the PO procedure bodies,
 back to the regular procedure bodies that call those PO procedures, would solve
 the portability problem but would not work functionally. There would be no
-guarantee that a call to :ada:`Put_Line` would immediately follow the execution
+guarantee that, during execution, the call to :ada:`Put_Line` would immediately
+follow the execution
 of the protected procedure called immediately before it in the source code.
 Hence the printed text might not reflect the current value of the :ada:`State`
 variable.
--- a/content/courses/ada-in-practice/chapters/raii.rst
+++ b/content/courses/ada-in-practice/chapters/raii.rst
@@ -26,17 +26,19 @@ Although concurrency is a prime example, the issue is general in nature. We
 will continue with the concurrency context for the sake of discussion.
 
 Like the classic *monitor* concept (:footcite:p:`1977:hansen`,
-:footcite:p:`1973:hansen`, :footcite:p:`1974:hoare`) on which they are based.
-(In the works cited, Hoare's contribution was equally important, but Hansen's
-contributions were reified in
-:wikipedia:`Concurrent Pascal <Concurrent_Pascal>`, a concrete programming
-language.), Ada defines a protected object (PO) as a concurrency construct that
+:footcite:p:`1973:hansen`, :footcite:p:`1974:hoare`) on which they are based,
+Ada defines a protected object (PO) as a concurrency construct that
 is higher-level and more robust than mutexes and semaphores. Those advantages
 accrue because the bodies of the protected operations are only responsible for
 implementing the functional requirements. The underlying run-time library is
 responsible for implementing the mutually exclusive access, and also thread
 management. As a result, the source code is much simpler and is robust even in
 the face of exceptions.
+
+(In the works cited above, Hoare's contribution was equally important, but
+Hansen's contributions were reified in
+:wikipedia:`Concurrent Pascal <Concurrent_Pascal>`, a concrete programming
+language.)
 
 However, a protected object is not always appropriate. Consider an existing
 sequential program that makes calls to visible procedures provided by a
--- a/content/courses/ada-in-practice/chapters/raii.rst
+++ b/content/courses/ada-in-practice/chapters/raii.rst
@@ -1,0 +1,480 @@
+.. _Ada_In_Practice_RAII:
+
+Resource Acquisition Is Initialization (RAII)
+=============================================
+
+.. include:: ../../global.txt
+
+Motivation
+----------
+
+In order for the expected semantics to be obtained, some types require clients
+to follow a specific protocol when calling the type's operations. Furthermore,
+failing to follow the protocol can cause system-wide ill effects.
+
+For example, in the context of competing concurrent threads accessing shared
+resources, concurrency abstractions such as mutexes provide the mutually
+exclusive access necessary to prevent race conditions. These mutex objects must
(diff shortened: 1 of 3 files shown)


gusthoff:
I've just made the changes you suggested. Just let me know if there's anything else I should improve here.

content/courses/ada-in-practice/chapters/raii.rst
```diff
@@ -0,0 +1,480 @@
...
+:wikipedia:`Concurrent Pascal <Concurrent_Pascal>`, a concrete programming
+language.), Ada defines a protected object (PO) as a concurrency construct that
+is higher-level and more robust than mutexes and semaphores. Those advantages
+accrue because the bodies of the protected operations implement only the
+functional requirements, with guaranteed mutually exclusive access to the
+protected data. The underlying implementation, i.e., the run-time library,
+provides the necessary exclusive access, and also thread management. As a
+result, the lock is guaranteed to be acquired, and likewise, guaranteed to be
+relinquished, even in the face of exceptions.
+
```
pat-rogers, line 41: Let's change the sentences that end the paragraph above (starting at line 35) to the following:
accrue because the bodies of the protected operations are only responsible for implementing the
functional requirements. The underlying run-time library is responsible for 
implementing the mutually exclusive access, and also thread management. As a
result, the source code is much simpler and is robust even in the face of exceptions.

content/courses/ada-in-practice/chapters/raii.rst
```diff
@@ -0,0 +1,480 @@
...
+We could put the PO in the package spec and then clients could call the
+protected object's operations directly, but changing all the clients could be
+expensive.
+
+However, this new design is not portable because the two :ada:`Threadsafe`
+protected procedure bodies both call a potentially blocking operation, in this
+case :ada:`Ada.Text_IO.Put_Line`. Erroneous execution is the result. It might
+work as intended when executed, or it might do something else, or, if detected,
+:ada:`Program_Error` will be raised. On a run-time library built on top of an
+operating system, it may work because the OS may provide thread locking
```
pat-rogers, line 157: "it may work because" => "it may work as intended because"

content/courses/ada-in-practice/chapters/raii.rst
```diff
@@ -0,0 +1,480 @@
...
+same PO could access the enclosed variables while the first caller is blocked,
+thus breaking the mutually exclusive access guarantee.
+
+Calling an I/O operation is not all that strange here, and those are not the
+only potentially blocking operation defined by the language.
+
+Note that moving the calls to :ada:`Put_Line` out of the PO procedure bodies,
+back to the regular procedure bodies that call those PO procedures, would solve
+the portability problem but would not work functionally. There would be no
+guarantee that a call to :ada:`Put_Line` would immediately follow the execution
```
pat-rogers, line 175: Let's change the above to:
"guarantee that, during execution, the call to :ada:`Put_Line` would immediately follow the execution"

gusthoff:
I've just finished those adaptations. If you have more suggestions, just let me know!

pat-rogers (dismissed):
Many thanks!

content/courses/ada-in-practice/chapters/raii.rst
```diff
@@ -0,0 +1,481 @@
...
+To define the :ada:`Lock_Manager` type, we declare it in a separate package as
+a tagged limited private type with a discriminant designating a
+:ada:`Mutual_Exclusion` object:
+
+.. code-block:: ada
+
+      type Lock_Manager (Lock : not null access Mutual_Exclusion) is
+         tagged limited private;
+
+We make it a limited type because copying doesn't make sense semantically for
```
swbaird, line 312: There are other good reasons for making the type limited - you can decide whether this point is worth mentioning.
For example, see AAARM (Annotated Ada Reference Manual) 7.6(27.a), where it says
    The goal of the above permissions is to allow typical dead assignment and dead variable removal
    algorithms to work for nonlimited controlled types.

Note the important word "nonlimited". For this application, you very much do not want to allow dead variable removal.
This is my only comment, other than to say that it all looks good.

pat-rogers: Let's add the following after that sentence:
In addition, during optimization the compiler is allowed to remove unreferenced objects of non-limited types. As you saw above in procedure :ada:`Op`, there will be no explicit references to the object :ada:`LM`, so making the type limited prevents that unwanted optimization.

gusthoff: I've just added that sentence.

gusthoff: @swbaird: if you haven't already, would you mind checking for grammar / typos as well?

pat-rogers:
> I've just added that sentence.
Thanks!!

content/courses/ada-in-practice/chapters/raii.rst
```diff
@@ -0,0 +1,486 @@
...
+order. Failure to acquire the mutex prior to accessing the shared resource
+leads to race conditions, and failure to release it can lead to deadlocks.
+Ensuring the mutex is released is complicated by the possibility of exceptions
+raised after the lock is acquired.
+
+Although concurrency is a prime example, the issue is general in nature. We
+will continue with the concurrency context for the sake of discussion.
+
+Like the classic *monitor* concept (:footcite:p:`1977:hansen`,
+:footcite:p:`1973:hansen`, :footcite:p:`1974:hoare`) on which they are based.
```
swbaird, line 29: Did you really want that period after "based".  That means the next sentence starts with a parenthesized note, which is odd.

content/courses/ada-in-practice/chapters/raii.rst
```diff
@@ -0,0 +1,486 @@
...
+manipulate that lock object. We use an access discriminant to designate that
+lock. By doing so, we decouple the new type from the specific lock, and thus
+from the application code. Otherwise, the new facility would not be reusable.
+
+The resulting relationship between the global shared lock and the local object
+will be as follows:
+
+.. code-block:: ada
+
+       Lock : Mutual_Exclusion;
```
swbaird, line 278: Do you want to declare Lock with the keyword "aliased", or is that too much detail for this rough sketch of the code? Later on when this design is refined, the word "aliased" is added. Either way is fine with me, but I'm raising the question.

content/courses/ada-in-practice/chapters/raii.rst
```diff
@@ -0,0 +1,486 @@
...
+    with Ada.Finalization;
+    with GNAT.Semaphores; use GNAT.Semaphores;
+
+    package Lock_Managers is
+
+      subtype Mutual_Exclusion is Binary_Semaphore
+        (Initially_Available => True,
+         Ceiling             => Default_Ceiling);
+
+       type Lock_Manager (Lock : not null access Mutual_Exclusion) is
```
swbaird, line 357: Why include the word "tagged"?  Maybe this is a fine idea, if we want to allow users to extend this type. OTOH, this seems like unnecessarily exposing an implementation detail. I'm just raising the question for you to consider; either answer is fine with me.

content/courses/ada-in-practice/chapters/raii.rst
```diff
@@ -0,0 +1,486 @@
...
+       procedure Operation_2 is
+          LM : Lock_Manager (Lock'Access) with Unreferenced;
+       begin
+          State := State - 1;  -- for example...
+          Put_Line ("State is now" & State'Img);
+       end Operation_2;
+
+    end P;
+
+The aspect :ada:`Unreferenced` tells the compiler that no references in the
```
swbaird, line 431: Do we want to mention that Unreferenced is a GNAT-defined aspect? In the context of this document, do we want to clarify what is portable vs. what is GNAT-defined? Even if we want to make this distinction, I think there is no need to say anything about uses of child units of GNAT, specifically GNAT.Semaphores.

content/courses/ada-in-practice/chapters/raii.rst
```diff
@@ -0,0 +1,486 @@
...
+type. But by the same token, the compiler will issue a warning if some explicit
+reference is present, perhaps added much later in the project lifetime.
+
+Pros
+----
+
+Race conditions are precluded, the client code is simpler than direct manual
+calls, and the code is robust, especially concerning exceptions. These
+advantages are significant, given the cost in engineering time to debug the
+errors this design prevents.
```
swbaird, line 445: If two pieces of state each have their own associated lock, then that introduces the possibility of deadlock. Two tasks each acquire one of the two and want to acquire the other. So that's an advantage of using one lock for different pieces of state even if they could be safely accessed concurrently.

content/courses/ada-in-practice/chapters/raii.rst
```diff
@@ -0,0 +1,486 @@
...
+Cons
+----
+
+The lock is global, so all calls go through it. Hence all calls are sequential,
+even if some could run concurrently. In the above example that's exactly as
+required, but in other situations it might be unnecessarily limiting.
+
+Compared to the manual call approach, the run-time cost for keeping track of
+objects to be finalized could be non-trivial. That's likely true in any
+language.
```
swbaird, line 457: No distinction is made between readonly access vs. readwrite. If we had a function that requires only readonly access to a piece of state, mutiple calls to that function could safely run concurrently. With this implementation, they won't. RIght?

swbaird (commented):
I only found one possible typo (see my first comment about a questionable period).

content/courses/ada-in-practice/chapters/raii.rst
pat-rogers: No, it should be a comma. Good catch!

content/courses/ada-in-practice/chapters/raii.rst
pat-rogers: I thought it was more detail than appropriate for that sketch.

content/courses/ada-in-practice/chapters/raii.rst
pat-rogers: I like to use tagged types when they're not too much trouble, on the theory that I don't know if someone will need to extend it in the future.

content/courses/ada-in-practice/chapters/raii.rst
pat-rogers: Right

PR #1260, opened 2025-09-16
Enhance the Cons section to say this is as type-safe as streams can be.

pat-rogers:
Stress that streams can contain values of multiple types by definition, so this is as type-safe as can be.

Do some minor formatting and polishing.

### DIFF
--- a/content/courses/ada-in-practice/chapters/streams_api_flexibility.rst
+++ b/content/courses/ada-in-practice/chapters/streams_api_flexibility.rst
@@ -18,7 +18,7 @@ to send and receive more complex data types. If so, how can the device driver
 support them?
 
 We certainly don't want the clients to do unchecked conversions everywhere.
-That's error-prone, because it prevents the compiler from checking the usage,
+That's error-prone, it prevents the compiler from checking the usage,
 and makes clients responsible for what should be an internal implementation
 detail.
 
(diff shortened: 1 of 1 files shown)


gusthoff (approved):

pat-rogers:
On 19-Sep-25 7:01 AM, Gustavo A. Hoffmann wrote:
> ***@***.**** commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In content/courses/ada-in-practice/chapters/streams_api_flexibility.rst 
> <https://github.com/AdaCore/learn/pull/1260#discussion_r2362664908>:
> 
>> @@ -18,7 +18,7 @@ to send and receive more complex data types. If so, how can the device driver
>   support them?
>   
>   We certainly don't want the clients to do unchecked conversions everywhere.
> -That's error-prone, because it prevents the compiler from checking the usage,
> +That's error-prone, it prevents the compiler from checking the usage,
> 
> Minor nitpick: I'd replace the first comma by a colon, i.e.:
> 
> ⬇️ Suggested change
> 
> -That's error-prone, it prevents the compiler from checking the usage,
> +That's error-prone: it prevents the compiler from checking the usage,
No, I disagree, that would have the same meaning as including the word 
"because" in the first version. It is error-prone in itself, and also 
prevents the compiler from checking for those errors...

PR #1371, opened 2026-06-27
New content: add section on ordinary fixed-point types to Numerics

gusthoff:


content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
-    scale factor: for a decimal fixed-point type, the scaling, given explicitly by
-    the type's :ada:`delta`, is always a power of ten.
+Ordinary fixed-point types are similar to decimal fixed-point types in that the
+values are, in effect, scaled integers. The difference between them is in the
+scale factor: for a
+:ref:`decimal fixed-point type <Adv_Ada_Decimal_Fixed_Point_Types>`, the
+*small* always equals its *delta*, which must be a power of ten.
+In contrast, for an ordinary fixed-point type, the scaling of the type's
+*small* is a power of two by default. Therefore, ordinary fixed-point
+types are sometimes called binary fixed-point types.
```
swbaird: This makes it sound like
    type T is delta 1.0 / 3.0 with Small => 1.0 / 3.0;
would be called a binary fixed-point type. That seems wrong.
How about "An ordinary fixed -point type with a *small* that is a power of two is sometimes called ...".
Also, talking about the scaling of the type's Small seems wrong; the Small value is not being scaled.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
 
-    In contrast, for an ordinary fixed-point type, the scaling is defined by the
-    type's :ada:`small`, which is derived from the specified :ada:`delta` and, by
-    default, is a power of two. Therefore, ordinary fixed-point types are sometimes
-    called binary fixed-point types.
+.. note::
+    Ordinary fixed-point types can be thought of being closer to the actual
```
swbaird: Again, this comment doesn't seem right if the small is not a power of two. Having defined the term "binary fixed point type" above, perhaps use it here?

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
-          delta <delta-value>
-          range <lower-bound> .. <upper-bound>;
+    type <type-name> is
+      delta <delta-value>
+      range <lower-bound> .. <upper-bound>;
 
-    By default the compiler will choose a scale factor, or :ada:`small`, that is a
-    power of 2 no greater than <delta-value>.
+By default the compiler will choose a scale factor, or :ada:`small`, that is a
+power of 2 no greater than <delta-value>.
```
swbaird, line 7345: Do we want to talk about what language allows (any power of two that is leq the delta) vs. what GNAT implements. Perhaps give a reference to https://docs.adacore.com/live/wave/gnat_rm/html/gnat_rm/gnat_rm/compatibility_and_porting_guide.html#writing-portable-fixed-point-declarations
That section also discusses the shaving rule, which seems worth mentioning if you haven't already talked about it elsewhere. RM 3.5.9(13) says "An ordinary_fixed_point_definition defines an ordinary fixed point type whose base range includes at least all multiples of small that are between the bounds specified in the real_range_specification. The base range of the type does not necessarily include the specified bounds themselves." That's where the shaving rule comes from.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
-    normalized range. When running the application, we notice that the upper
-    bound is close to one, but not exact one. This is a typical effect of
-    fixed-point data types |mdash| you can find more details in this discussion
-    about the :wikipedia:`Q format <Q_(number_format)>`.
+There are actually two formats: one of them counts the sign bit, the other one
+does not. In this course, we use the latter, i.e. the format that doesn't count
+the sign bit.
 
-    We may also rewrite this code with an exact type definition:
+The Q format consists of two numbers: the number of bits for the integer part
```
swbaird: My first reaction is that perhaps we don't want to mention Q format at all. Or mention it very briefly. For example, after defining an example type
     Del : constant := 2.0 ** -8;
     Lo : constant := -(2.0 ** 7);
     type Q7_8 is delta Del range Lo .. -(Lo + Del) with Small => Del;
, we could add a comment
    For readers familiar with Q numeric format (see <wikipedia reference>), this would be a Q7.8 format type.
    The identifier chosen for the type (i.e., "Q7_8") reflects its Q format; the same is true for many
    types declared in subsequent examples in this document.

Q format can't describe Ada fixed point types  where the small is not a power of two, but it also doesn't work for cases where the delta is a power of two that is larger than one:
     type Coarse is delta 2.0 ... with Small => 2.0;

But I don't feel strongly. If you think it is useful to discuss Q format, I don't object. It does describe the cases that come up most commonly in practice.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
+
+Small and delta
+~~~~~~~~~~~~~~~
+
+As we already mentioned
+:ref:`in a previous section <Adv_Ada_Fixed_Point_Types_Small_Delta>`, the small
+of a decimal type is always equal to the delta that we specified. However, for
+ordinary fixed-point types, this doesn't have to be the case |mdash| and if we
+select a delta that is not a power of two (i.e. the typical machine
+representation), the compiler will choose a *small* that is the largest power
```
swbaird: This GNAT behavior is common for other implementations, but is not required. See previous reference to GNAT documentation regarding portability issues for fixed point types.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
+≈ 2.44×10\ :sup:`-4`) is smaller than :ada:`Angle'Delta` (1/3600
+≈ 2.78×10\ :sup:`-4`): the compiler picked the largest power of two
+not exceeding the *delta*. This means stored angle values are rounded
+to the nearest multiple of 2\ :sup:`-12`, which may not coincide with
+exact multiples of 1/3600.
+
+By contrast, for :ada:`Angle_2`, we use :ada:`with Small => Angle_Delta` to
+force *small* = *delta*, so every multiple of 1/3600 is representable
+exactly. Note that Ada allows non-power-of-two *small* values via the
+:ada:`Small` aspect, but not all implementations are required to support
```
swbaird: odd phrasing. it sounds like some implementations are required to do something while other implementations are not.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
+|        0.25 |     =   8,192 * 2\ :sup:`-15` |
++-------------+-------------------------------+
+|        0.50 |     =  16,384 * 2\ :sup:`-15` |
++-------------+-------------------------------+
+
+.. admonition:: For further reading...
+
+    As you might have expected, two fixed-point types with the same size can
+    have different machine representations. Again, the actual integer value is
+    based solely on the type's *delta*, and not the type's size.
```
swbaird: delta => small ?

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
+         2#0.111_1111_1111_1111_1111_1111_1111_1111#;
+       V_15 := TQ15 (V_31);
+       Show_Vars;
+    end Show_Fixed_Point_Conversions;
+
+
+As we've done before, we can use an
+:ref:`overlay <Adv_Ada_Address_Aspect_Overlay>` to uncover the actual integer
+values stored on the machine when assigning values to objects of fixed-point
+type. For example:
```
swbaird: Does it go without saying that when using overlays we have to be careful to make sure types match up? We discussed all this (e.g., alignment) before in the context of another document, so perhaps no need to repeat. But perhaps mention that the issue exists and give a reference? Your call; I don't feel strongly.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
+notation |mdash| for instance, :ada:`0.25` as ``0.25000`` |mdash| with
+enough fractional digits to reflect the type's *small*. Converting to
+:ada:`Float` first and using :ada:`Float'Image`, on the other hand,
+produces the floating-point representation in exponential notation, such
+as ``2.50000E-01``.
+
+Let's focus on the :ada:`Show (0.1)` call. Note that the value :ada:`0.1` isn't
+a multiple of the *small* of :ada:`TQ15`. In fact, the value 0.1 is rounded to
+the nearest representable value when we assign it. For this reason,
+:ada:`TQ15'Image` shows ``0.09998`` rather than ``0.10000``.
```
swbaird: And similarly for the Float'Image call too. The loss of the exact value 0.1 takes place when Show is called, so there is nothing to be done inside of Show to recover it. Perhaps this isn't worth mentioning, but if you think there is a chance that readers might think that converting to Float inside Show will somehow magically restore the exact 0.1 value then that would be worth clarifying. Your call.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
+                 & TQ15_Quarter'Last'Image);
+    end Show_Derived_Fixed_Point_Ranges;
+
+In this example, :ada:`TQ15_Half` and :ada:`TQ15_Quarter` are both derived
+from :ada:`TQ15`. For :ada:`TQ15_Half`, we constrain the range to -0.5 to
+(0.5 - *small*). For :ada:`TQ15_Quarter`, we constrain it further to -0.25
+to (0.25 - *small*).
+
+
+Range of fixed-point subtypes
```
swbaird, line 8582: I agree with the decision to say nothing at all about reduced accuracy subtypes (RM J.3)

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
+         range -1.0 .. 1.0 - D;
+
+       function Sat_Add (A, B : Sample)
+                         return Sample is
+          pragma Suppress (Overflow_Check);
+          pragma Suppress (Range_Check);
+
+          C : constant Sample := A + B;
+       begin
+          --  Two same-signed operands whose sum
```
swbaird: No, no, no!  RM 11.5(26) says: "If a given check has been suppressed, and the corresponding error situation occurs, the execution of the program is erroneous." If adding a Suppress pragma has any impact at all on the behavior of a program, then that means program execution is erroneous. If the suppressed check would never fail, then behavior is unaffected. If it would have failed, then execution is erroneous. Relying on check suppression to do something particular that the user had in mind (e.g., wrapping) is a common error.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
+    end Show_Saturating;
+
+This version produces the same results as the exception-based one |mdash|
+:ada:`Sat_Add (0.5, 0.75)` and :ada:`Sat_Mul (-1.0, -1.0)` both saturate to
+:ada:`Sample'Last`, while :ada:`Sat_Add (0.5, 0.25)` returns 0.75 unchanged
+|mdash| but no exception is ever raised or handled.
+
+The trick relies on :ada:`Sample` being a Q15 type that fills its 16-bit
+machine word exactly. Suppressing the overflow check lets an out-of-range
+operation *wrap around* at the type bounds, just as the underlying machine
```
swbaird: For other things that the suppressing an overflow check lets an out-of-range operation do, see the reference to "nasal demons" in the wikipedia article on "undefined behavior". I like the movie "The Wizard of Oz" at least as much as most folks, but let's avoid winged monkeys.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7292,149 +7297,2983 @@ a complex financial application can be significant and, therefore, it might be
...
+roughly 0.55, the term reaches :math:`\pm 1.0` and the assignment to
+:ada:`D.Z1` raises :ada:`Constraint_Error` (or wraps around silently when range
+checks are suppressed). The half-scale input is therefore not arbitrary
+|mdash| it is what keeps the feedback term inside the type's range.
+
+We can make the filter robust for *any* input by giving the feedback path the
+headroom it needs. In fact, when designing DSP algorithms |mdash| especially
+when targeting fixed-point types |mdash| we have to make sure that the result
+is **always** in the expected type range. Using headroom is common practice to
+guarantee this is always true.
```
swbaird, line 10456: In talking about how much headroom is available, would it be worth mentioning System.Fine_Delta? Maybe not. If you are given either bounds or a power-of-two small, then typically find_delta can be used to statically get a maximal value for the other element (i.e. if given bounds, then compute small via F_D; if given small, then similarly compute bounds). this isn't strictly guaranteed to work (and it won't work for enormous values), but it will work in practice (not just for GNAT); and if it doesn't work, it will fail at compile-time. But perhaps this is going too far afield.

swbaird (commented):
Not approving because of check suppression misuse.

content/courses/advanced-ada/parts/data_types/numerics.rst
gusthoff: I think it's useful to discuss the Q format. However, as a compromise, I've just moved most of the subsection into an admonition (which the reader may skip), so that core is now small.

content/courses/advanced-ada/parts/data_types/numerics.rst
gusthoff: OK, you're right! I'll use a different approach.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7313,17 +7313,21 @@ values are, in effect, scaled integers. The difference between them is in the
 scale factor: for a
 :ref:`decimal fixed-point type <Adv_Ada_Decimal_Fixed_Point_Types>`, the
 *small* always equals its *delta*, which must be a power of ten.
-In contrast, for an ordinary fixed-point type, the scaling of the type's
-*small* is a power of two by default. Therefore, ordinary fixed-point
-types are sometimes called binary fixed-point types.
+In contrast, an ordinary fixed-point type's *small* is a power of two by
+default. However, unlike for decimal fixed-point types |mdash| which always
+use a power of ten |mdash| this isn't mandatory: if we use the :ada:`Small`
```
swbaird: Perhaps "if we use the Small aspect" => "by specifying the Small aspect" ?
Not a big deal.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7706,9 +7720,9 @@ Subtypes of ordinary fixed-point types
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 A subtype of an ordinary fixed-point type has the same *delta* and *small* as
-its parent type; the only constraint allowed in a subtype declaration is a
-range constraint. Unlike derived types, subtype objects are directly
-interchangeable with the parent type |mdash| no explicit conversion is needed.
+its parent type; the only non-obsolescent constraint allowed in a subtype
```
swbaird: It might be better to not mention obsolescent precision constraints at all. Whatever you think is best.

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -9465,10 +9661,127 @@ cannot represent exactly:
...
+the exact mathematical sum 1.222222...
+
+.. admonition:: In the Ada Reference Manual
+
+    - :arm22:`G.2.3 Model of Fixed Point Arithmetic <G-2-3>`
+
+
+.. _Adv_Ada_Ordinary_Fixed_Point_System_Fine_Delta:
+
+System.Fine_Delta
```
swbaird, line 9687: Nice new section!

swbaird (dismissed):
Looks good. I particularly like the new Fine_Delta section, and avoiding the use of unsafe check suppression.

content/courses/advanced-ada/parts/data_types/numerics.rst
gusthoff: As a compromise, I moved  the comment to an admonition ("for further reading").

content/courses/advanced-ada/parts/data_types/numerics.rst
gusthoff: Thanks! :-)

content/courses/advanced-ada/parts/data_types/numerics.rst
```diff
@@ -7722,9 +7722,15 @@ Subtypes of ordinary fixed-point types
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 A subtype of an ordinary fixed-point type has the same *delta* and *small* as
-its parent type; the only non-obsolescent constraint allowed in a subtype
-declaration is a range constraint. A subtype declaration may also include
-a delta constraint, but, as mentioned above, that's an obsolescent feature.
+its parent type; the only constraint allowed in a subtype declaration is a
```
swbaird: This sentence is immediately contradicted by the next sentence. Perhaps add a parenthetical "(without using obsolescent language features)" or some similar qualifier.

swbaird (dismissed):
One minor point. I'll approve now to let you decide whether a correction is needed.

content/courses/advanced-ada/parts/data_types/numerics.rst
gusthoff: You're totally right, thanks for the hint!

swbaird (approved):

